PR #2, opened 2021-07-17
allow access to metadata on participant struct

ArthurHlt:
For now, user can't access to metadata store in a participant, the only way to get it is to use listener.

This make following code possible:

```go
import (
  lksdk "github.com/livekit/server-sdk-go"
)

func main() {
  host := "<host>"
  apiKey := "api-key"
  apiSecret := "api-secret"
  roomName := "myroom"
  identity := "botuser"
	room, err := lksdk.ConnectToRoom(host, lksdk.ConnectInfo{
		APIKey:              apiKey,
	})
	if err != nil {
		panic(err)
	}

  room.Callback.OnParticipantConnected = func(participant *lksdk.RemoteParticipant) {
     metadata = participant.Metadata()
  }

}
```

### DIFF
--- a/participant.go
+++ b/participant.go
@@ -56,6 +56,10 @@ func (p *baseParticipant) Identity() string {
 	return p.identity
 }
 
+func (p *baseParticipant) Metadata() string {
+	return p.metadata
+}
+
 func (p *baseParticipant) IsSpeaking() bool {
 	val := p.isSpeaking.Load()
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
good catch! we missed exposing metadata previously.

ArthurHlt:
Thanks for livekit btw ! You've all did a great work
and also thanks for fast merging my PR

PR #4, opened 2021-07-17
Fix callback for participant disconnected

ArthurHlt:
This is a small fix for the callback when participant disconnect, it actually use `OnParticipantConnected` instead of `OnParticipantDisconnected`


davidzhao (approved):
really good catch. thanks for fixing! btw, do you mind if I added you to the livekit org? I really appreciate the PRs!

ArthurHlt:
@davidzhao It would be a pleasure, just I will maybe not have too much time for working on livekit

davidzhao:
Definitely, not expecting anything, just making it easier to keep track of contributors. Cheers!

PR #7, opened 2021-07-24
handle lossy data packet messages which fix speakers update

ArthurHlt:
Livekit server use two differents protocol for sending speakers update: https://github.com/livekit/livekit-server/blob/master/pkg/rtc/room.go#L519-L525

It's seems that's server send lossy data packet for speakers update to room from go sdk. But they are not handled so I started implementation for reading it in the same for as `SignalClient`. This currently handle speakers update and I don't really know what should be handled too. 

This also include a small fix on `Participan.setIsSpeaking` which was not update speaking if the state of speaking state change. It should be inverted obviously

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -5,6 +5,8 @@ import (
 	"time"
 
 	"github.com/pion/webrtc/v3"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
 
 	livekit "github.com/livekit/server-sdk-go/proto"
 )
@@ -146,6 +148,7 @@ func (e *RTCEngine) configure(res *livekit.JoinResponse) error {
 	if err != nil {
 		return err
 	}
+	e.lossyDC.OnMessage(e.handleLossyDataPacket)
 	e.reliableDC, err = e.publisher.PeerConnection().CreateDataChannel(reliableDataChannelName, &webrtc.DataChannelInit{
 		Ordered: &trueVal,
 	})
@@ -218,6 +221,29 @@ func (e *RTCEngine) waitUntilConnected() error {
 
 func (e *RTCEngine) handleLocalTrackPublished(res *livekit.TrackPublishedResponse) {
 	e.trackPublishedChan <- res
+}
+
+func (e *RTCEngine) handleLossyDataPacket(msg webrtc.DataChannelMessage) {
+	packet, err := e.ReadDataPacket(msg)
+	if err != nil {
+		return
+	}
+	switch msg := packet.Value.(type) {
+	case *livekit.DataPacket_Speaker:
+		if e.OnActiveSpeakersChanged != nil {
+			e.OnActiveSpeakersChanged(msg.Speaker.Speakers)
+		}
+	}
+}
+
+func (e *RTCEngine) ReadDataPacket(msg webrtc.DataChannelMessage) (*livekit.DataPacket, error) {
+	dataPacket := &livekit.DataPacket{}
+	if msg.IsString {
+		err := protojson.Unmarshal(msg.Data, dataPacket)
+		return dataPacket, err
+	}
+	err := proto.Unmarshal(msg.Data, dataPacket)
+	return dataPacket, err
 }
 
 func (e *RTCEngine) negotiate() error {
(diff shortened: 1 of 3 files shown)


participant.go
```diff
@@ -89,7 +89,7 @@ func (p *baseParticipant) setAudioLevel(level float32) {
 
 func (p *baseParticipant) setIsSpeaking(speaking bool) {
 	lastSpeaking := p.IsSpeaking()
-	if speaking != lastSpeaking {
+	if speaking == lastSpeaking {
```
davidzhao, line 92: great catch!

engine.go
```diff
@@ -220,6 +223,29 @@ func (e *RTCEngine) handleLocalTrackPublished(res *livekit.TrackPublishedRespons
...
+	}
+	switch msg := packet.Value.(type) {
+	case *livekit.DataPacket_Speaker:
+		if e.OnActiveSpeakersChanged != nil {
+			e.OnActiveSpeakersChanged(msg.Speaker.Speakers)
+		}
+	}
+}
+
+func (e *RTCEngine) ReadDataPacket(msg webrtc.DataChannelMessage) (*livekit.DataPacket, error) {
```
davidzhao, line 239: this is probably better to be package private.

```suggestion
func (e *RTCEngine) readDataPacket(msg webrtc.DataChannelMessage) (*livekit.DataPacket, error) {
```

engine.go
```diff
@@ -146,6 +148,7 @@ func (e *RTCEngine) configure(res *livekit.JoinResponse) error {
 	if err != nil {
 		return err
 	}
+	e.lossyDC.OnMessage(e.handleLossyDataPacket)
```
davidzhao, line 151: we should hook up reliableDC to this as well.

davidzhao (commented):
Hey Arthur, great catch here on the isSpeaking bug!

For handling data packets, we kind of missed it as well. and I think your change works! We use data channels to allow clients to publish arbitrary data into the room, and it's their choice if it's reliable or not. Here's the [JS implementation](https://github.com/livekit/client-sdk-js/blob/main/src/room/RTCEngine.ts#L271) on handling. Here's [the docs](https://docs.livekit.io/guides/room/data#data-message-from-participants) on it in case you are curious how it's used on other platforms.

engine.go
davidzhao: and also rename the function. to the client, a message is received is the same regardless. it's the sender's choice whether they'd like reliable delivery or not.

participant.go
ArthurHlt: thanks :)

engine.go
ArthurHlt: I've made it private thus

engine.go
ArthurHlt: I've handle it for realiable too and renamed to `handleDataPacket`

ArthurHlt:
You will see also  that I've handled the receive data callback which was not handled  ;)

engine.go
```diff
@@ -220,6 +225,33 @@ func (e *RTCEngine) handleLocalTrackPublished(res *livekit.TrackPublishedRespons
...
+	packet, err := e.readDataPacket(msg)
+	if err != nil {
+		return
+	}
+	switch msg := packet.Value.(type) {
+	case *livekit.DataPacket_Speaker:
+		if e.OnActiveSpeakersChanged != nil {
+			e.OnActiveSpeakersChanged(msg.Speaker.Speakers)
+		}
+	case *livekit.DataPacket_User:
```
ArthurHlt, line 238: add reading for user packets

room.go
```diff
@@ -134,6 +135,18 @@ func (r *Room) handleDisconnect() {
 	r.engine.Close()
 }
 
+func (r *Room) handleDataReceived(userPacket *livekit.UserPacket) {
```
ArthurHlt, line 138: add handling in room for using previously made callback `OnDataReceived`

engine.go
```diff
@@ -220,6 +225,33 @@ func (e *RTCEngine) handleLocalTrackPublished(res *livekit.TrackPublishedRespons
...
+		if e.OnDataReceived != nil {
+			e.OnDataReceived(msg.User)
+		}
+	}
+}
+
+func (e *RTCEngine) readDataPacket(msg webrtc.DataChannelMessage) (*livekit.DataPacket, error) {
+	dataPacket := &livekit.DataPacket{}
+	if msg.IsString {
```
davidzhao, line 247: technically speaking, this can be removed. datapackets are always encoded in proto to ensure the lowest overhead

davidzhao (approved):
looks great.

PR #8, opened 2021-07-25
fix version of go module twirp

ArthurHlt:
When protobuf has been regenerated and commited on https://github.com/livekit/server-sdk-go/pull/5 it did not include required go module update for twirp. This PR simply update module consequently.


davidzhao (approved):
looks good!

PR #9, opened 2021-07-27
fix race connection when handling active speakers

ArthurHlt:
When an active speaker is disconnecting in some conditions you can get this kind of error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc23ca5]

goroutine 98 [running]:
github.com/livekit/server-sdk-go.(*RemoteParticipant).setAudioLevel(0x0, 0xc03e000000)
	<autogenerated>:1 +0x5
github.com/livekit/server-sdk-go.(*Room).handleActiveSpeakerChange(0xc0004113b0, 0xc00000ea68, 0x1, 0x1)
	server-sdk-go/room.go:196 +0x46c
```

Probable condition is when setting active speakers and have delete from map just before/after it. GC should have marked already the element to be deleted.

This is because `var p Participant` is an interface which can be not be nil when checking but when nil when execute fonction on it. It is really hard to explain and this link: https://getstream.io/blog/fixing-the-billion-dollar-mistake-in-go-by-borrowing-from-rust/ explain better than me on what might occurred.

When setting address directly to var it never happens again

### DIFF
--- a/room.go
+++ b/room.go
@@ -179,15 +179,15 @@ func (r *Room) handleActiveSpeakerChange(speakers []*livekit.SpeakerInfo) {
 	seenSids := make(map[string]bool)
 	for _, s := range speakers {
 		seenSids[s.Sid] = true
-		var p Participant
-
 		if s.Sid == r.LocalParticipant.sid {
-			p = r.LocalParticipant
+			r.LocalParticipant.setAudioLevel(s.Level)
+			r.LocalParticipant.setIsSpeaking(true)
+			activeSpeakers = append(activeSpeakers, r.LocalParticipant)
 		} else {
-			p = r.GetParticipant(s.Sid)
-		}
-
-		if p != nil {
+			p := r.GetParticipant(s.Sid)
+			if p == nil {
+				continue
+			}
 			p.setAudioLevel(s.Level)
 			p.setIsSpeaking(true)
 			activeSpeakers = append(activeSpeakers, p)


davidzhao:
Thanks for the fix. Interface vars can indeed be nil, but stops being nil once a concrete type is assigned to it (even if the value is nil). This has caught me a couple of times in the past. This snippet illustrates the problem:

https://play.golang.org/p/tf752MCv6Pz

ArthurHlt:
Btw, I have another suggestion. We could use a [sync.Map](https://pkg.go.dev/sync#Map) instead of a map and `RWMutex` this make things faster (in our case map is read a lot and sync.Map perform on this) a and easier as you don't need to remember to Lock and Unlock.

I would like to do so for all `tracks` in *Participant* and `participants` in *Room*. I would also make them private instead of public in struct, there is already method for both of them but it will break current UX.

**Note**: I've tried to fix this issue with a [sync.Map](https://pkg.go.dev/sync#Map) but it did not fix this issue. (But probably other race condition)

ArthurHlt:
@davidzhao thanks for sharing, this is actually the second time I had this issue in my golang programmer life :)

davidzhao:
Re: Map suggestion. It's true that sync.Map is faster for a read-heavy workload, however there are a couple of considerations:

* sync.Map doesn't enforce types (at least until Go supports generics). This makes it slightly more error prone. Making them private would help to compensate for some of that; though how do users iterate through them?
* RWMutex is fairly fast, how many times do you need to access these structures?

I'm open to the change,  but want to make sure we get the interface right.

ArthurHlt:
- There is already a method GetParticipants which give a slice, simply reconstruct a slice like this would works. But yes we can't definitely let access to sync.Map without control over it
- It is not really about call lock or unlock but more when, as in some listener lock as been called before. If we take the function `GetParticipants` inside this listener, you will be blocked as lock has been already called and this function call it too

davidzhao:
it's a fair point regarding the locks being tricky to work with. That's a good reason to simplify. as long as we ensure that the GetX(id), and GetXs() methods are available, I'm good with that update.

PR #10, opened 2021-07-29
Use sync map instead of map with lock

ArthurHlt:
This PR replace map for participants and tracks to sync map for ease usage.

Discussion can be found here: https://github.com/livekit/server-sdk-go/pull/9#issuecomment-887751966

Breaking change for user inside:
- list of participants in a room can't be accessed anymore by getting `Participants` attribute. Instead, user should always use method `GetParticipants()` or `GetParticipant(sid string)`

### DIFF
--- a/participant.go
+++ b/participant.go
@@ -18,6 +18,7 @@ type Participant interface {
 	Metadata() string
 }
 
+// map[string]TrackPublication
 type baseParticipant struct {
 	sid        string
 	identity   string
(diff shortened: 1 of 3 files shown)


frostbyte73 (approved):
Lgtm!

davidzhao:
hey @ArthurHlt thanks for these solid PRs that you've been opening. [Join us on slack](https://join.slack.com/t/livekit-users/shared_invite/zt-rrdy5abr-5pZ1wW8pXEkiQxBzFiXPUg) if you are up for it; we have a channel for contributors to discuss idea and changes.

PR #12, opened 2021-08-04
Recording apis

frostbyte73:


recordingclient.go
```diff
@@ -0,0 +1,53 @@
...
+func (c *RecordingServiceClient) EndRecording(ctx context.Context, req *livekit.EndRecordingRequest) (*livekit.RecordingResponse, error) {
+	ctx, err := c.withAuth(ctx, auth.VideoGrant{RoomAdmin: true})
+	if err != nil {
+		return nil, err
+	}
+	return c.RecordingService.EndRecording(ctx, req)
+}
+
+func (c *RecordingServiceClient) withAuth(ctx context.Context, grant auth.VideoGrant) (context.Context, error) {
+	at := auth.NewAccessToken(c.apiKey, c.apiSecret)
```
davidzhao, line 45: would be great to use the same helper vs copying code. can we move this, apikey/secret to an authhelper class that we can composite?

recordingclient.go
```diff
@@ -0,0 +1,53 @@
...
+	client := livekit.NewRecordingServiceProtobufClient(url, &http.Client{})
+	return &RecordingServiceClient{
+		RecordingService: client,
+		apiKey:           apiKey,
+		apiSecret:        secretKey,
+	}
+}
+
+func (c *RecordingServiceClient) StartRecording(ctx context.Context, req *livekit.StartRecordingRequest) (*livekit.RecordingResponse, error) {
+	ctx, err := c.withAuth(ctx, auth.VideoGrant{RoomAdmin: true})
```
davidzhao, line 29: similar comment as in server. currently roomAdmin requires a room. let's chat how we'll want to treat this permission.

davidzhao (approved):
looks good!

PR #29, opened 2022-01-10
can set participants name in ConnectInfo

mheers:


davidzhao:
lg! thanks for the PR

PR #30, opened 2022-01-11
LocalSampleTrack: calculate / interpolate duration when buffer duration is invalid

mkrufky:
Calculate / interpolate duration value when buffer duration is invalid

We're seeing a video stream come through Gstreamer with timestamped buffers without duration values.
These streams fail to play through Livekit.  This fixes it by calculating the duration when missing / invalid.

Signed-off-by: Michael Krufky <michael.krufky@livecontrol.io>

### DIFF
--- a/localsampletrack.go
+++ b/localsampletrack.go
@@ -37,6 +37,7 @@ type LocalSampleTrack struct {
 	bound        uint32
 	lock         sync.RWMutex
 	audioLevelID uint8
+	lastTS       time.Time
 
 	cancelWrite func()
 	provider    SampleProvider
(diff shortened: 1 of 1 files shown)


localsampletrack.go
```diff
@@ -209,6 +210,12 @@ func (s *LocalSampleTrack) WriteSample(sample media.Sample, opts *SampleWriteOpt
 		s.sequencer.NextSequenceNumber()
 	}
 
+	// calculate / interpolate duration when supplied duration is invalid
+	if sample.Duration.Nanoseconds() < 0 {
```
davidzhao, line 214: should we be doing this when duration is set to 0 as well? Do you know what would cause duration to be negative?

davidzhao (commented):
Looks good!

localsampletrack.go
mkrufky: The stream causing this issue has `-1` duration set.  It might be OK if we check for `<= 0` but it isn't necessary for this specific issue.

PR #32, opened 2022-01-19
Introduce `recordbot` to record stream directly

dwirya:
The current official recording solution is [livekit-recorder](https://github.com/livekit/livekit-recorder), but it wasn't suitable for our use case to save individual video track. The existing recorder composites the videos even if it's a single video stream, and it consumes considerable resource which does not scale well for our needs.

Our solution is to read from the RTP packets and save them to file without any processing in between. As an add-on, we were able to upload to S3 afterwards, although there needs more documentation for setting up the IAM role and attaching to the EC2 instance or similar instances.


vladlensb:
Nice work! Thank you

pkg/recordbot/recorder.go
```diff
@@ -0,0 +1,159 @@
...
+	// Hooks provide access to customising aspects of the recorder, such as uploading the file after recording
+	hooks RecorderHooks
+}
+
+type RecorderHooks struct {
+	// The recorder will pass the filename that it was instantiated with, to the hook
+	UploadFile func(filename string) error
+}
+
+func NewRecorder(filename string, codec webrtc.RTPCodecParameters, hooks RecorderHooks) (Recorder, error) {
```
davidzhao, line 48: sounds like the idea for this recorder is a single track? should we rename the type/constructor to indicate as such? maybe `NewSingleTrackRecorder`

it'd be great to one day add a recorder that would mux audio and video together into MP4

pkg/recordbot/recorder.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func createMediaWriter(fileName string, mimeType string) (media.Writer, error) {
+	if mimeType == webrtc.MimeTypeVP8 || mimeType == webrtc.MimeTypeVP9 {
+		return ivfwriter.New(fileName)
+	}
+	if mimeType == webrtc.MimeTypeH264 {
+		return h264writer.New(fileName)
+	}
+	if mimeType == webrtc.MimeTypeG722 ||
```
davidzhao, line 68: could a ogg container contain PCMA/PCMU/G722?

pkg/recordbot/recorder.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	return nil, ErrUnsupportedCodec
+}
+
+func (r *recorder) Start(track *webrtc.TrackRemote) {
+	r.wg.Add(1)
+	go r.record(track)
+}
+
+func (r *recorder) record(track *webrtc.TrackRemote) {
```
davidzhao, line 84: I think this loop could be written a bit differently to be easier to read.

You can keep these struct members:
* closedChan chan struct{}
* done chan struct{}

`done` has a single purpose of notifying that a user has requested stop. In this case it's more idiomatic to close it since that'll notify all pending readers.

```
var err error
defer func() {
    // handle error
    if err != nil {
    }
    // move cleanup routine in here
    r.cleanup()
    if r.hooks.UploadFile != nil ...
    
    // signal Stop that this is done
    close(r.closeChan)
}()
for {
    select {
    case <-done:
        return
    default:
        // Read RTP stream
        var packet rtp.Packet
	packet, _, err = track.ReadRTP()
	if err != nil {
		return
	}

	// Write to file
	err = writer.WriteRTP(packet)
	if err != nil {
		return
	}
    }
}
```

pkg/recordbot/uploader_s3.go
```diff
@@ -0,0 +1,71 @@
...
+
+type s3Uploader struct {
+	config S3Config
+	client s3iface.S3API
+}
+
+var ErrEmptyS3BucketName = errors.New("empty S3 bucket name")
+var ErrEmptyS3ACL = errors.New("empty S3 ACL")
+
+func NewS3Uploader(config S3Config) (Uploader, error) {
```
davidzhao, line 32: I'm conflicted if this belongs in livekit-sdk. It pulls in aws-sdk-go and is quite platform specific. wdyt about having users handle this on their own? by passing in an appropriate uploader interface.

davidzhao (commented):
Really good work! I'm excited to have this functionality in the Go SDK. Thank you for making this available to all.

dwirya:
Thanks for your comprehensive feedback @davidzhao ! I’ll work on this over the weekend and update you next week :)

pkg/recordbot/uploader_s3.go
dwirya: Yeah I thought so too. The idea was to provide a plug-and-play functionality to upload to S3. But I realised this is better left for the API consumer as this introduces unnecessary dependencies. I’ll delete the S3 code and tidy up the dependencies

pkg/recordbot/recorder.go
dwirya: That’s a good idea which I didn’t think of. Thanks for the suggestion! I’ll probably need to define an interface for the 2 recorders

pkg/recordbot/recordbot.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	// Stop recording
+	rec.Stop()
+
+	// Remove recorder from list
+	delete(bot.recorders, recorderID)
+}
+
+func generateRecorderID(identity string, codecType webrtc.RTPCodecType) (string, error) {
```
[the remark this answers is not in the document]
dwirya: I’m guessing one use case where this fails is if they’re doing screen sharing or track publishing. Either way, SID sounds good to me!

dwirya:
Hi @davidzhao , I've incorporated the feedback in your review in my latest commit. Please let me know if you have further feedback ! The dual track recorder is not something our company is interested in, so I'll be working on it in my free time. ETA is about 2 weeks within my comfort zone. Apologies if it'll take slightly long considering there's a growing interest in this.

pkg/recordbot/recorder.go
dwirya: Hi @davidzhao, I didn't implement the struct members as I was unsure why this is the recommended practice. Perhaps you can clarify this ?
- Why is using an empty struct better than a boolean ?
- Why do we need both `closedChan` and `done` ? I would assume `done` is sufficient just for tracking the `stop` signal.

davidzhao: struct{} is preferred as it only indicates the type, but consumes 0 bytes. The main point I'm trying to make is closing the channel instead of sending a value to it. Closing ensures two things:
1. it'll not block, you can always close a channel. writing to a channel would block if it's unbuffered
2. close would notify all readers that they can no longer expect to get values from it

`closeChan` is used to allow `Stop()` to block, until post processing is finished. For example, if you were to Stop the recorder, then immediately read the file from S3, it may error out if Stop returned before the write operation is complete

davidzhao:
> Hi @davidzhao , I've incorporated the feedback in your review in my latest commit. Please let me know if you have further feedback ! The dual track recorder is not something our company is interested in, so I'll be working on it in my free time. ETA is about 2 weeks within my comfort zone. Apologies if it'll take slightly long considering there's a growing interest in this.

no rush at all :) you are helping a lot of people with this implementation. I appreciate it!

pkg/recordbot/recorder.go
dwirya: I think I understand. This means I can use either buffered or unbuffered channels for `done` and I shouldn’t run into deadlock right ? But for `closedChan`, it will have to be an unbuffered channel and I’ll have something like `<- closedChan` in `Stop()` to block the cleanup process until it’s done.

dwirya:
Hi @davidzhao , I've pushed my change regarding the empty-struct channels. Thanks for your feedback, overall the code looks cleaner now without WaitGroup. I tested the changes locally on my recordbot service, and it still works. Still can't fully grasp how it all works together (it's so elegant yet feels like black magic to me 😅), but it's coming together. Let me know if you need further changes.

davidzhao:
hey @dwirya, running the tests here and I'm seeing it hang. Could you also give me write permissions to this repo? I have a cleanup change that's easiest if I just push to the branch.

dwirya:
@davidzhao thanks for testing it out! I haven't encountered it but that's because I haven't considered all cases. I'll have a look once you've pushed the changes; I also have some cleanup changes for the recorder.

pkg/recordbot/recordbot.go
```diff
@@ -9,7 +9,6 @@ import (
 	"github.com/pion/webrtc/v3"
 )
 
-var ErrRecorderNotFound = errors.New("recorder not found")
```
davidzhao: I didn't see this error used anywhere.

pkg/recordbot/recorder.go
```diff
@@ -111,7 +112,8 @@ func (r *recorder) record(track *webrtc.TrackRemote) {
 			return
 		default:
 			// Read RTP stream
-			packet, _, err := track.ReadRTP()
+			var packet *rtp.Packet
+			packet, _, err = track.ReadRTP()
```
davidzhao, line 116: avoid shadowing err declared in the outer context

pkg/recordbot/recorder_test.go
```diff
@@ -41,65 +41,65 @@ func cleanupTestRecorder() {
 func TestCanCreateH264Writer(t *testing.T) {
 	mimeType := webrtc.MimeTypeH264
 	writer, err := createMediaWriter("test.mp4", mimeType)
-	assert.NotNil(t, writer)
-	assert.NoError(t, err)
+	require.NotNil(t, writer)
```
davidzhao, line 44: `require` in testify works like `assert` in other languages. where it halts tests upon the first require failure.

elhe26:
Any updates?

dwirya:
Currently there are 2 deliverables for this feature:
1. Docker service that can manage recording when hitting `/recordings/start` and `/recordings/stop` given the room name. Generated file name can be configurable via YAML, but this will take some time. The default is `{identity}_{start-timestamp}`. It’s also uploadable to S3 by specifying the region, bucket and optionally, directory. This is what our team is using in production.
2. Integration into LiveKit Go SDK. This is a much more flexible approach, which is what this PR is about. Currently, we’re waiting on a jitter buffer implementation to handle out-of-order packets from LiveKit’s team and we can close this PR.

dwirya:
@elhe26 can you clarify what you meant by saving records using S3 ? Are you referring to uploading it to S3 without saving it to a local file ?

dwirya:
I think I can get the Docker service up by next week if you wanna try it out @elhe26 . Currently I’m doing some major rewrites to make the recording hopefully more stable

elhe26:
@dwirya We're planning on using DO spaces to save the streams. A general approach regarding documentation, where you can specify a general configuration (region, bucket, etc...) for S3 compatible providers (DO, Wasabi, Blackblaze, etc...). 

Is there a way to start and stop recordings programmatically using docker?

dwirya:
@elhe26 I haven't looked into DO and the likes, but the S3 uploading is using `aws-sdk-go-v2`. However, seeing https://stackoverflow.com/questions/58622647/how-to-upload-file-into-wasabi-bucket-with-s3-api-with-node-js, it seems that it's very possible: I just need to add an environment variable such as `AWS_S3_ENDPOINT` or so.

The way to start and stop recording programmatically is to hit the `/recordings/start` and `/recordings/stop` POST endpoints with the payload `{ "room": $room_name }`

elhe26:
@dwirya, are you managing the container using k8s or as a standalone?  I'm trying to wrap my head here. You're using server-sdk-go inside a container, right?

dwirya:
I wanted to use K8, but ditched it in favour of standalone. Essentially, I'm running it in an EC2 instance. The container image is a Go application that uses server-sdk-go. I don't think K8 is a good fit because we need to store the recordings locally before uploading to S3. I'm still new to K8, but I think K8 is more suitable for stateless applications such as API server.

Another idea I've been toying around with is streaming directly to S3 (which @davidzhao recommended), skipping the need to save files locally, which makes K8 a good choice. However, I can only think of the possibility for raw files (video `.h264`, `.ivf` and audio `.ogg`) which I don't think browsers support. For any container files (`.webm`, `.mp4`), we should be using `ffmpeg` or `gstreamer` instead of rolling our own media writer, but using those require saving the files to local disk.

If you have further suggestions for deployment, we can take this offline on Slack. But presently, using standalone works for my team.

pkg/recordbot/recorder.go
```diff
@@ -0,0 +1,136 @@
...
+	if mimeType == webrtc.MimeTypeH264 {
+		return h264writer.New(fileName)
+	}
+	if mimeType == webrtc.MimeTypeG722 ||
+		mimeType == webrtc.MimeTypePCMA ||
+		mimeType == webrtc.MimeTypePCMU ||
+		mimeType == webrtc.MimeTypeOpus {
+		// The values for oggwriter can be found at
+		// https://github.com/pion/webrtc/blob/master/examples/save-to-disk/main.go#L53
+		return oggwriter.New(fileName, 48000, 0)
```
rrampage, line 73: channelCount should be either 1 or 2, or `.ogg` file gets created with muted audio. The linked example uses OggWriter with 2 channels.

dwirya: Hi @rrampage , thanks for the catch! I wouldn't know about it.

dwirya:
I'll be doing some major rewrites, mainly around the API and also incorporating sample buffer. The reviews here will be implemented in the commits for the rewrite

cnderrauber:
Have created a PR #40 for jitter buffer integrated for record track.

dwirya:
Closing as it's stale and there were significant changes in the SDK. Will reopen a new PR

PR #34, opened 2022-02-02
returns error early

mheers:


davidzhao:
thank you for fixing this! this was a miss on my part.

PR #35, opened 2022-02-03
Remove explicit setting DTLSRole to client

davidzhao:
Integrated Pion WebRTC https://github.com/pion/webrtc/pull/2110


boks1971 (approved):
There is no go.sum change?

davidzhao:
> There is no go.sum change?

oops, forgot to run go mod tidy.

PR #40, opened 2022-02-14
add example for save track to file

cnderrauber:
integrate samplebuilder from https://github.com/jech/samplebuilder/blob/master/samplebuilder.go.
modify it for pli send and raw rtp packets read.

add example for save track to file.(ivfwriter use frame count as PTS, that might cause video played in a incorrect framerate(fast or slow)


boks1971 (approved):
Did not check the sample builder code. Assumed it is good at the source :-)

pkg/samplebuilder/samplebuilder.go
```diff
@@ -0,0 +1,519 @@
+// This SampleBuilder is copied from https://github.com/jech/samplebuilder.
+// Modify it to integrate with media writer of pion
```
davidzhao, line 2: nit

```suggestion
// Modified to integrate with Pion's media.Writer
```

examples/filesaver/main.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func trackSubscribed(track *webrtc.TrackRemote, publication *lksdk.RemoteTrackPublication, rp *lksdk.RemoteParticipant) {
+	fileName := fmt.Sprintf("%s-%s", rp.Identity(), track.ID())
+	fmt.Println("write track to file ", fileName)
+	NewTrackWriter(track, rp.WritePLI, fileName)
+}
+
+const (
+	maxLate = 200
```
davidzhao, line 66: can we increase this to 2s or something very high? since for recording it doesn't really care how behind it is

examples/filesaver/main.go
```diff
@@ -0,0 +1,129 @@
...
+func NewTrackWriter(track *webrtc.TrackRemote, pliWriter lksdk.PLIWriter, fileName string) (*TrackWriter, error) {
+	var (
+		sb     *samplebuilder.SampleBuilder
+		writer media.Writer
+		err    error
+	)
+	switch {
+	case strings.EqualFold(track.Codec().MimeType, "video/vp8"):
+		sb = samplebuilder.New(maxLate, &codecs.VP8Packet{}, track.Codec().ClockRate, samplebuilder.WithPacketDroppedHandler(func() {
+			pliWriter(track.SSRC())
```
davidzhao, line 84: in addition to packets being dropped, we should write it periodically to ensure keyframes are generated every N seconds, for seekability

davidzhao (commented):
looks good!

examples/filesaver/main.go
cnderrauber: this is a packet sequence latency, 200 packets means 4s for audio. 2s for medium quality video. ideally we can optimize this latency to seconds later

examples/filesaver/main.go
cnderrauber: for playback enhancement(seekability, reduce file size and others), we'd better to transcode after record completed. send pli periodically will affect other participant in the room (higher bitrate may cause higher packet loss)

examples/filesaver/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	case strings.EqualFold(track.Codec().MimeType, "video/h264"):
+		sb = samplebuilder.New(maxLate, &codecs.H264Packet{}, track.Codec().ClockRate, samplebuilder.WithPacketDroppedHandler(func() {
+			pliWriter(track.SSRC())
+		}))
+		writer, err = h264writer.New(fileName + ".h264")
+
+	case strings.EqualFold(track.Codec().MimeType, "audio/opus"):
+		sb = samplebuilder.New(maxLate, &codecs.OpusPacket{}, track.Codec().ClockRate)
+		writer, err = oggwriter.New(fileName+".ogg", 48000, 2)
```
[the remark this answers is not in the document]
cnderrauber: For opus packet, it has a header to indicate these(samplerate, stereo/mono...) , so these will not affect playback.
But your suggestion has better readability, will change it.

examples/filesaver/main.go
cnderrauber: use two latency for audio/video now

Issue #49, opened 2022-03-07
first PublishData always results in an error


qwertzui11:
Hi! 

Thx for your great work with livekit. It's a really awesome framework and very easy to use! 

However I'm getting an error when sending some data

```go
room.Callback.OnParticipantConnected = func(p *lksdk.RemoteParticipant) {
  if err := room.LocalParticipant.PublishData([]byte("hello"), livekit.DataPacket_RELIABLE, []string{p.SID()}); err != nil {
    log.Printf("could not publish data to participant '%s', err: '%v'", p.SID(), err)
  }
}
```
After a few SECONDS this call results in the error
```
could not publish data to participant 'PA_kCS2E6r2YV3c', err: 'could not connect after timeout'
```
Immediately after that log entry, the following log entry occurs:
```
"level"=0 "msg"="successfully set publisher answer"
```

After that log entry I'm able to successfully send data messages. It seems to me that the first call to `PublishData` blocks some logic.

If I try to send with a go routine like this
```go
	room.Callback.OnParticipantConnected = func(p *lksdk.RemoteParticipant) {
		log.Printf("OnParticipantConnected '%s'", p.Identity())
		go func() {
			time.Sleep(time.Second * 1)
			sendData(p)
			time.Sleep(time.Second * 1)
			sendData(p)
		}()
	}
```
the first send results in the error 
```
could not publish data to participant 'PA_73nWKR3avB7X', err: 'io: read/write on closed pipe'
```
the second send succeeds.

In conclusion: the first `PublishData` always fails. Afterwards it works.

I'm not sure if it matters but the other participant is using the [js-sdk](https://github.com/livekit/client-sdk-js)

Any help would be greatly appreciated  :smile:

cnderrauber:
since publisher is not establish until you publish something(mediatrack or data), so you publish data in OnParticipantConnected will cause publisher pc to connect and block the signal msg process loop, that's why you get timeout.

If you publish data at goroutine, the publisher pc will connected then open the data channel, the publishdata method just wait publish pc connected but data channel may still opening, so your first pub might get closed pipe.

There are two points to improve:
1. we might need make callback in goroutine to avoid user do something block our signal loop @davidzhao 
2. create datachannel after publish pc connected, that might solve your first pub failed problem

qwertzui11:
> 2. create datachannel after publish pc connected, that might solve your first pub failed problem

[...]

PR #50, opened 2022-03-08
ensure datachannel ready before publish data

cnderrauber:


cnderrauber:
also make room callbacks to be async. fix #49

engine.go
```diff
@@ -243,12 +243,12 @@ func (e *RTCEngine) waitUntilConnected() error {
 	}
 }
 
-func (e *RTCEngine) ensurePublisherConnected() error {
+func (e *RTCEngine) ensurePublisherConnected(dataReady bool) error {
```
boks1971, line 246: Maybe rename this to `checkDataReady`. Just `dataReady` feels like it is a signal about readiness of data channel. Or maybe `ignoreDataReady`.

boks1971 (approved):

engine.go
cnderrauber: done

cnderrauber:
I'm not sure, both pion and lk-server use OpsQueue in internal logic, will observe this part, can change to use OpsQueue if we get performance/timing problem with go func()

PR #51, opened 2022-03-11
feat: Introduce `trackrecorder` package to save track to file

dwirya:
Continuation from https://github.com/livekit/server-sdk-go/pull/32 as there were multiple significant changes to the SDK. The track recorder now uses context as it is preferred to cancel goroutines. The previous implementation used channels, which is fine, but there were instances where blocking occurs. Goroutine seems more idiomatic as well. 

The package also exposes some helpers, such as to determine extension based on codec. I can write an example of how to use the new package (if approved), but not sure if I should make a new example or improve on existing one in `examples/filesaver`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 .idea/
 docs/
 bin/
+.vscode/
--- a/examples/filesaver/main.go
+++ b/examples/filesaver/main.go
@@ -1,23 +1,20 @@
 package main
 
 import (
-	"errors"
+	"context"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"os/signal"
-	"strings"
 	"syscall"
+	"time"
 
-	"github.com/pion/rtp/codecs"
-	"github.com/pion/webrtc/v3"
-	"github.com/pion/webrtc/v3/pkg/media"
-	"github.com/pion/webrtc/v3/pkg/media/h264writer"
-	"github.com/pion/webrtc/v3/pkg/media/ivfwriter"
-	"github.com/pion/webrtc/v3/pkg/media/oggwriter"
-
+	"github.com/livekit/protocol/auth"
 	lksdk "github.com/livekit/server-sdk-go"
 	"github.com/livekit/server-sdk-go/pkg/samplebuilder"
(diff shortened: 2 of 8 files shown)


dwirya:
To run the tests in VSCode, go to Preferences (UI) -> Workspace Tab -> Go: Test Env Vars -> Edit in settings.json and paste the following:
```
{
    "go.testEnvVars": {
        "LIVEKIT_URL": "ws://localhost:7880",
        "LIVEKIT_API_KEY": "...",
        "LIVEKIT_API_SECRET": "..."
    }
}
```

cnderrauber:
improve filesaver would be better I think, they do same things (save/record track to disk)

dwirya:
I've updated the example and tested locally. The only catch is the participant has to enter the room before we start recording.

dwirya:
Can I get you to review the PR @cnderrauber ?

elhe26:
Any updates?

davidzhao:
@dwirya since track recording will be provided with egress, it probably doesn't make sense to add this into the Go SDK. We'd want to keep it light-weight.

Do you want to make it available in your own repo? We can link to it from the Go SDK as additional resources for folks.

dwirya:
Yeah I agree. If I’m understanding this correctly, the current `recorder` package will be imported from our repository instead. If so, is this the right manual to do so: https://go.dev/doc/modules/publishing ?

dwirya:
@elhe26  can I know your use case to be specific ? I’m assuming you’d want to record per participant (track-composite). If you can afford time, I’d recommend waiting for the egress or trying out our recorder at https://github.com/cloudgroundcontrol/livekit-recorder (track-composite works, but doesn’t do transcoding. H264 is put in MP4 & VP8 in WebM). 

I personally don’t think maintaining your own recorder is a good idea (including this PR) as there’s much more saving the file. The simple way is to use Pion’s API, but I’m seeing issues with it at work that I’m leaning towards GStreamer now.
If it’s not an option (such as wanting to produce live m3u8 file), then maintaining your own recorder makes sense.

dwirya:
Closed as it's recommended to use a recorder service such as `livekit-egress` rather than SDK implementation

PR #64, opened 2022-05-20
Publish from stdin

dwirya:
- FileSampleProvider refactored to MediaSampleProvider, which reads from `io.ReadCloser` interface to handle multiple egress types
- Expose 2 methods to create track: `NewLocalFileTrack` (which does the same thing as before) and `NewLocalReaderTrack` (which is more low-level; accepts `io.ReadCloser`)
- Publishing from stdin is as simple as `NewLocalReaderTrack(os.Stdin, ...)`


davidzhao (approved):
great work!

davidzhao (approved):
looks good!

PR #72, opened 2022-06-24
update to protocol 8, solve data race

cnderrauber:


room.go
```diff
@@ -43,14 +44,20 @@ func WithAutoSubscribe(val bool) ConnectOption {
...
 
 type Room struct {
 	engine           *RTCEngine
-	SID              string
-	Name             string
+	sid              string
+	name             string
 	LocalParticipant *LocalParticipant
-	Callback         *RoomCallback
+	callback         *RoomCallback
```
davidzhao, line 60: This will break existing applications. We need to update documentation showing them how to use this, especially using `NewRoomCallback()` and overriding fields before passing it in.

room.go
```diff
@@ -65,9 +72,9 @@ func CreateRoom() *Room {
 	r := &Room{
 		engine:       engine,
 		participants: &sync.Map{},
-		Callback:     NewRoomCallback(),
```
davidzhao, line 68: since Callback is a private field now, we should take it in as an argument to `CreateRoom`

room.go
```diff
@@ -33,6 +33,7 @@ type ConnectInfo struct {
 type ConnectParams struct {
 	AutoSubscribe bool
 	Reconnect     bool
+	Callback      *RoomCallback
```
davidzhao, line 36: hmm, now I'm thinking it might be a better idea to take this in as an explicit argument in `ConnectToRoom`, users cannot leave this blank.

davidzhao (commented):
change looks good. but we should update README and other docs prior to merging. We also need to update livekit-cli to use the updated APIs.

davidzhao (commented):
change looks good. but we should update README and other docs prior to merging. We also need to update livekit-cli to use the updated APIs.

room.go
```diff
@@ -60,14 +61,15 @@ type Room struct {
 }
 
 // CreateRoom can be used to update callbacks before calling Join
-func CreateRoom() *Room {
+func CreateRoom(callback *RoomCallback) *Room {
```
cnderrauber, line 64: @davidzhao agree explicit parameter is better that most case user should pass a callback here, but can also pass a nil callback if he don't care any events (just connect and publish a data msg).

davidzhao: sounds like a plan!

davidzhao (approved):

PR #88, opened 2022-07-13
Update remote participant info on a recreate

boks1971:
A new remote participant could be created when handling OnTrack.
This could happen with fast connect.
Remote participant could also be added from join response.
Update participant info when re-creating.

Also, use map with lock for remote participants as there were
range calls on sync.Map in several places.

Before this, there were track subscription failures due to fast connect.

### DIFF
--- a/localparticipant.go
+++ b/localparticipant.go
@@ -270,10 +270,6 @@ func (p *LocalParticipant) GetPublisherPeerConnection() *webrtc.PeerConnection {
 
 func (p *LocalParticipant) updateInfo(info *livekit.ParticipantInfo) {
 	p.baseParticipant.updateInfo(info, p)
-	p.lock.Lock()
-	p.sid = info.Sid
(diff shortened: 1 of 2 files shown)


localparticipant.go
```diff
@@ -270,10 +270,6 @@ func (p *LocalParticipant) GetPublisherPeerConnection() *webrtc.PeerConnection {
 
 func (p *LocalParticipant) updateInfo(info *livekit.ParticipantInfo) {
 	p.baseParticipant.updateInfo(info, p)
-	p.lock.Lock()
-	p.sid = info.Sid
-	p.identity = info.Identity
-	p.lock.Unlock()
```
boks1971, line 276: This is in `baseParticipant` ☝️ and `updateInfo` of `baseParticipant` takes care of updating those fields.

cnderrauber (approved):

Issue #111, opened 2022-08-03
Automatic reconnection fails and never triggers Disconnect event


EmrysMyrddin:
I have upgraded livekit sdk and the new reconnection mechanism seems broken, or I don't understand how it is supposed to work.

In my case, when a websocket failure happens, there is 2 reconnection attempts in a row failing for different reasons:

1. when the websockets is closed due to network error, livekit fails to resume connection:
```
2022-08-02T11:38:02+02:00 2022/08/02 09:38:02 "msg"="error with read worker" "error"="read tcp 10.42.12.145:33880->104.21.31.106:443: read: connection reset by peer"
2022-08-02T11:38:02+02:00 2022/08/02 09:38:02 "level"=0 "msg"="engine.go handleDisconnect"
2022-08-02T11:38:02+02:00 2022/08/02 09:38:02 [Livekit] Reconnecting to LiveKit
2022-08-02T11:38:02+02:00 2022/08/02 09:38:02 "level"=0 "msg"="engine.go resumeConnection"
2022-08-02T11:38:02+02:00 2022/08/02 09:38:02 "level"=0 "msg"="signalclient.go Join" "url"="wss://livekit.clawmaster.app" "token"="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NTk0MzM0MTksImlzcyI6Im1haW4tc2VjcmV0LWtleSIsIm5iZiI6MTY1OTQzMjgxOSwic3ViIjoibnRzLTAtMTAiLCJ2aWRlbyI6eyJyb29tIjoibnRzLTAtMTAiLCJyb29tSm9pbiI6dHJ1ZX19.16Vvu6CCu_aEgZRPR8uAl5KA8sxiEqvgs2tHVysy6GU"
[...]

PR #120, opened 2022-08-17
fix resume 

cnderrauber:
For resume connection, should not expect JoinResponse. 
Add log for resume and restart connection


boks1971 (approved):

davidzhao (approved):
good fix. do we have a test that can validate that this works?

cnderrauber:
> good fix. do we have a test that can validate that this works?

we have a unittest for resume by simulate, but it passed by fallback to  restart after resume failed.

cnderrauber:
This should fix #111 , test failed for the server in docker, will merge this after livekit-server docker image updated

PR #134, opened 2022-10-06
Revert "1.17 compat, revert #128"

boks1971:
Reverts livekit/server-sdk-go#133

Now using https://github.com/livekit/mediatransportutil


go.mod
```diff
@@ -1,14 +1,15 @@
 module github.com/livekit/server-sdk-go
 
-go 1.17
+go 1.18
```
davidzhao, line 3: Is this required? doing so would remove 1.17 compatibility, which we want to avoid for the SDK

davidzhao (commented):
nice work!

go.mod
boks1971: Should not be needed. Will check.

davidzhao (approved):

PR #146, opened 2022-10-28
set ice transport policy from server response

cnderrauber:


magefile.go
```diff
@@ -33,14 +34,86 @@ const (
...
+log_level: debug
+logging:
+    pion_level: info
+`
+	parameters := []string{
+		`run`, `-e`, `LIVEKIT_KEYS=` + testApiKey, `-e`, `LIVEKIT_CONFIG=` + string(confString),
+		`-d`, `--rm`, `-p7880:7880`, `-p7881:7881`, `-p7882:7882/udp`, `-p3478:3478/udp`,
+	}
+
+	for p := 30000; p <= 30020; p++ {
```
cnderrauber, line 108: docker on mac has problem with expose range ports

integration_test.go
```diff
@@ -159,3 +162,55 @@ func TestResume(t *testing.T) {
...
+
+		iceTransport := dtlsTransport.ICETransport()
+		if iceTransport == nil {
+			return nil, errors.New("no ICE transport")
+		}
+
+		return iceTransport.GetSelectedCandidatePair()
+	}
+
+	require.Eventually(t, func() bool { return reconnected.Load() && pub.engine.ensurePublisherConnected(true) == nil }, 5*time.Second, 100*time.Millisecond)
```
boks1971, line 207: Not an issue, but trying to understand. Is there a need to check for `reconnected` again here? It is checked in the previous `require.Eventually`.

boks1971 (approved):

integration_test.go
cnderrauber: Redundant code during debug, thanks!

PR #155, opened 2022-11-30
Makes livekit.RoomService unexported to avoid confusion

itzloop:
Hi,
Since methods of `RoomServiceClient`, in `server-sdk-go`, are adding authentication, exporting `RoomService` would make it confusing to use. 

If i call `client.RoomService.ListRooms(context.Background(), &livekit.ListRoomsRequest{})` i would get an authentication error, However calling `client.ListRooms(context.Background(), &livekit.ListRoomsRequest{})` will work. 

This issue exists for other clients as well and if you are agree with me i would be happy to make this change to all the clients.

Thanks.

### DIFF
--- a/egressclient.go
+++ b/egressclient.go
@@ -9,7 +9,7 @@ import (
 )
 
 type EgressClient struct {
-	livekit.Egress
+	egressClient livekit.Egress
 	authBase
 }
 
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
make sense. thanks for the PR!

davidzhao:
@itzloop do you mind signing our [CLA](https://cla-assistant.io/livekit/server-sdk-go?pullRequest=155)? I can merge it in right after.

itzloop:
Sorry for the delay.
I will sign it right after i made a change to other clients as well.

itzloop:
I signed the CLA and also applied the change to `EgressClient` and `IngressClient`.

davidzhao:
Thanks @itzloop!

PR #169, opened 2023-01-10
return the lk api error message on signal connection failures

paulwe:
when the singalling connection fails the error currently contains the http error message (ie bad handshake) which isn't very useful. instead return the response body which contains the api error (ie authentication failed because etc...)


cnderrauber (approved):
Looks good!

davidzhao (approved):
oh wow, TIL.

PR #175, opened 2023-01-15
Do not attempt to reconnect until fully connected

### LINKED ISSUE #156: Unexpected behavior of the ConnectToRoom function
Hi there!

### Problem
I am using server-sdk-go to publish a stream to a livekit server. When I call the `lksdk.ConnectToRoom` function and the connection happens through a turn server, then the first time I can’t connect (the logs show `ICE failed`) and, accordingly, the function returns `nil` and `error`, but under the hood there is a reconnect and after two or more attempts the logs show `ICE Connected`, but I can't publish the stream because the `ConnectToRoom` function returned `nil`. 

### Issue
It seems to me that the API of the `lksdk.ConnectToRoom` function is a bit unexpected due to what is happening under the hood.

### version
`github.com/livekit/server-sdk-go v1.0.4`

Thanks for your time!

davidzhao:
fixes #156

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -33,6 +33,7 @@ type RTCEngine struct {
 	lossyDCSub         *webrtc.DataChannel
 	trackPublishedChan chan *livekit.TrackPublishedResponse
 	subscriberPrimary  bool
+	hasConnected       atomic.Bool
 	hasPublish         atomic.Bool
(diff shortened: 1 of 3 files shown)


cnderrauber (approved):
looks good!

PR #201, opened 2023-03-29
adding SimulateSpeakerUpdate and SimulateMigration scenarios

real-danm:
This adds `SimulateSpeakerUpdate` and `SimulateMigration` scenarios to Room.

Set `SimulateSpeakerUpdateInterval = 5` so no api change is needed.

### DIFF
--- a/room.go
+++ b/room.go
@@ -22,7 +22,6 @@ const (
 	SimulateForceTCP
 	SimulateForceTLS
 	SimulateSpeakerUpdate
-	SimulateMigration
 
 	SimulateSpeakerUpdateInterval = 5
 )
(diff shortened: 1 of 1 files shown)


boks1971 (approved):

cnderrauber (approved):

room.go
```diff
@@ -545,6 +549,28 @@ func (r *Room) Simulate(scenario SimulateScenario) {
...
+			Message: &livekit.SignalRequest_Simulate{
+				Simulate: &livekit.SimulateScenario{
+					Scenario: &livekit.SimulateScenario_Migration{
+						Migration: true,
+					},
+				},
+			},
+		}
+		r.engine.client.SendRequest(req)
+		r.engine.client.OnLeave(&livekit.LeaveRequest{CanReconnect: true, Reason: livekit.DisconnectReason_CLIENT_INITIATED})
```
cnderrauber, line 573: Don't need call OnLeave here, as the server side will close signal connection at the migration case. And I think go-sdk don't support migration since pion don't support dtls restart.

PR #209, opened 2023-04-15
correctly handle LocalParticipant on update

theomonnom:
This caused all remote events ( TrackPublished, ParticipantConnected, TrackSubscribed, ... ) to be called for local resources.
Also it can be a breaking change for our users because `len(room.GetParticipant())` is now decremented by 1


davidzhao (approved):
great fix!

PR #210, opened 2023-04-15
correctly read opus packets duration

### LINKED ISSUE #192: Live "Opus" Audio
Our team has been working to resolve a long standing issue when it came to getting audio to sync, we have a slack conversation here: https://livekit-users.slack.com/archives/C01KZ61HZRR/p1670006468765049

Finally decided to take a look at what the SDK was doing and found that the issue we though was related to `gstreamer` and `tcpclientsrc` was actually caused by a few lines here: https://github.com/livekit/server-sdk-go/blob/main/readersampleprovider.go#L213

By removing the lines that were calculating `Duration` and simply always returning: `sample.Duration = defaultOpusFrameDuration` I am consistently getting immediate live audio & it works for files & sockets.

theomonnom:
This may fix #192 
Also the oggreader should correctly read the next opus packet instead of the page.
I will implement that when I have the time

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -17,7 +17,6 @@ import (
 const (
 	// defaults to 30 fps
 	defaultH264FrameDuration = 33 * time.Millisecond
-	defaultOpusFrameDuration = 20 * time.Millisecond
 )
 
 // ReaderSampleProvider provides samples by reading from an io.ReadCloser implementation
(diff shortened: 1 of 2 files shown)


readersampleprovider.go
```diff
@@ -202,18 +200,19 @@ func (p *ReaderSampleProvider) NextSample() (media.Sample, error) {
 		sample.Duration = time.Duration(p.ivfTimebase*float64(delta)*1000) * time.Millisecond
 		p.lastTimestamp = header.Timestamp
 	case webrtc.MimeTypeOpus:
-		pageData, pageHeader, err := p.oggreader.ParseNextPage()
+		// TODO(theomonnom): Read next OpusPacket instead of the next page
+		data, _, err := p.oggreader.ParseNextPage()
```
davidzhao, line 204: does ParseOpusPacketDuration rely on the fact that oggreader returns a packet, instead of a page?

utils.go
```diff
@@ -7,9 +7,60 @@ import (
...
+	"time"
 )
 
+var (
+	ErrInvalidOpusPacket = errors.New("invalid opus packet")
+)
+
+// Parse the duration of a an OpusPacket
+// https://www.rfc-editor.org/rfc/rfc6716#section-3.1
+func ParseOpusPacketDuration(data []byte) (time.Duration, error) {
```
davidzhao, line 21: 🔥 🔥

davidzhao: Awesome work here!

readersampleprovider.go
theomonnom: This is already working if the page contains only one opus packet.
But this isn't the right thing to do, we should correctly read the pages & segments inside the ogg container.

davidzhao (approved):

PR #219, opened 2023-04-29
More granular errors (actually only one more).

boks1971:
Main goal of this is to distinguish whether connection to the controller (signal channel) failed or if that succeeded, but could not get a media node for the connection.

In the current scheme of things, if websocket dial fails, a simple http.Get is done on `/rtc/validate` endpoint to understand the error better.

Just adding one more error return type.
I think if the error is `ErrCannotConnectSignal`, that is the one case where the media node could not be found.

Also logging the websocket dial error and HTTP response as that method can return the HTTP response (if any) along with error `ErrBadHandshake` for things like redirect, auth etc (https://github.com/gorilla/websocket/blob/af47554f343b4675b30172ac301638d350db34a5/client.go#L149). Want to observe what that returns and check if we can do an early return for dial failures. NOTE: Things could change between the web socket dial and the second http.Get. So, it would be nice to see if we can make a decision purely based on websocket dial return value.

For reviewers: the idea here is to not trip media self health checks on controller Dial failures.

### DIFF
--- a/signalclient.go
+++ b/signalclient.go
@@ -81,18 +81,21 @@ func (c *SignalClient) Join(urlPrefix string, token string, params *ConnectParam
 	}
 
 	header := newHeaderWithToken(token)
-	conn, _, err := websocket.DefaultDialer.Dial(u.String(), header)
+	conn, hresp, err := websocket.DefaultDialer.Dial(u.String(), header)
 	if err != nil {
+		logger.Errorw("error establishing signal connection", err, "httpResponse", hresp)
 		// use validate endpoint to get the actual error
 		validateSuffix := strings.Replace(urlSuffix, "/rtc", "/rtc/validate", 1)
 
-		res, err := http.Get(ToHttpURL(urlPrefix) + validateSuffix)
-		if err != nil || res.StatusCode == http.StatusOK {
+		hresp, err := http.Get(ToHttpURL(urlPrefix) + validateSuffix)
+		if err != nil {
+			return nil, ErrCannotDialSignal
+		} else if hresp.StatusCode == http.StatusOK {
 			// no specific errors to return if validate succeeds
 			return nil, ErrCannotConnectSignal
 		} else {
 			var errString string
-			switch res.StatusCode {
+			switch hresp.StatusCode {
 			case http.StatusUnauthorized:
 				errString = "unauthorized: "
 			case http.StatusNotFound:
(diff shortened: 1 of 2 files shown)


signalclient.go
```diff
@@ -81,26 +81,29 @@ func (c *SignalClient) Join(urlPrefix string, token string, params *ConnectParam
...
 		validateSuffix := strings.Replace(urlSuffix, "/rtc", "/rtc/validate", 1)
 
-		res, err := http.Get(ToHttpURL(urlPrefix) + validateSuffix)
-		if err != nil || res.StatusCode == http.StatusOK {
+		hresp, err := http.Get(ToHttpURL(urlPrefix) + validateSuffix)
+		if err != nil {
+			return nil, ErrCannotDialSignal
+		} else if hresp.StatusCode == http.StatusOK {
 			// no specific errors to return if validate succeeds
 			return nil, ErrCannotConnectSignal
```
boks1971, line 95: I am thinking this is the error for which the media self health should fail. This errors means, we were able to connect to the controller and do a validation. The validation was good, yet could not connect. That means, the media node cannot be picked/contacted.

Like I mentioned in the PR commit notes, I think if we can determine from WebSocket Dial return itself whether it is a dial failure or not, it would be great. Logging for now. We can potentially return errors during WebSocket dial and use those directly rather than second step.

matkam (approved):
This is a good first step. We'll see if we end up needing more error information on the initial WebSocket connection.

real-danm (approved):

signalclient.go
```diff
@@ -81,26 +81,29 @@ func (c *SignalClient) Join(urlPrefix string, token string, params *ConnectParam
 	}
 
 	header := newHeaderWithToken(token)
-	conn, _, err := websocket.DefaultDialer.Dial(u.String(), header)
+	conn, hresp, err := websocket.DefaultDialer.Dial(u.String(), header)
 	if err != nil {
+		logger.Errorw("error establishing signal connection", err, "httpResponse", hresp)
```
davidzhao, line 86: did you mean to log the entire response obj? or just the code?

I think when error is returned, response would might be nil

davidzhao (approved):

signalclient.go
boks1971: I intended to log as much as the response to get more information. So, yeah, will have to dial that back once I have some data. Gorilla documentation says that if there is a hanshake failure, it can return an error with a HTTP response non-nil (https://github.com/gorilla/websocket/blob/af47554f343b4675b30172ac301638d350db34a5/client.go#L149). That's why I added it. I want to see if there is a clear indication of Dial error vs application error.

Speaking of that, let me add a log for the validate end point get error also if any to check what that returns.

PR #223, opened 2023-05-09
Revert "correctly read opus packets duration (#210)"

frostbyte73:
#210 breaks the opus reader for ogg files


theomonnom (approved):
ouch, yes I think I need to move the oggreader from Kitt for this PR to be compatible

davidzhao (approved):

Issue #240, opened 2023-05-24
Republish track on full restart can't work as expectedly 


cnderrauber:
Two things need to be fixed:
In full restarting, the sdk doesn't save the entire publish option so it will republish track with track name only, other options like resolution, source, dtx is lost.
`LocalSampleTrack` can't support republishing, its unbind will clear provider information will cause next bind failed after republish succeeded.

PR #241, opened 2023-05-25
Fix republish don't work on full restart #240

cnderrauber:
The `LocalSampleTrack` will call SampleProvider.OnUnbind and cause it close the reader, if sdk republish the track on full reconnect, the republish track will failed on SampleProvider.Bind as the reader has been already closed. 
To fix the issue and keep API don't break already exists application, the PR add a `ReaderTrackDisableAutoClose` Option to SamplePrivoider which tells sdk don't close the reader when track unbinded, so the reader can continue read sample after track republished, also add `Close` method to the `LocalSampleTrack` for application to clean resource if `ReaderTrackDisableAutoClose` is enabled.
If application don't enable `ReaderTrackDisableAutoClose`, sdk will keep its behavior like before that clean reader on unbind and don't support republish on full reconnect.

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -43,6 +43,8 @@ type ReaderSampleProvider struct {
 	// for ogg
 	oggreader   *oggreader.OggReader
 	lastGranule uint64
+
+	dontCloseOnUnbind bool
 }
 
 type ReaderSampleProviderOption func(*ReaderSampleProvider)
@@ -68,6 +70,12 @@ func ReaderTrackWithOnWriteComplete(f func()) func(provider *ReaderSampleProvide
 func ReaderTrackWithRTCPHandler(f func(rtcp.Packet)) func(provider *ReaderSampleProvider) {
 	return func(provider *ReaderSampleProvider) {
 		provider.trackOpts = append(provider.trackOpts, WithRTCPHandler(f))
+	}
+}
+
+func ReaderTrackDisableAutoClose() func(provider *ReaderSampleProvider) {
+	return func(provider *ReaderSampleProvider) {
+		provider.dontCloseOnUnbind = true
 	}
 }
 
@@ -137,6 +145,11 @@ func NewLocalReaderTrack(in io.ReadCloser, mime string, options ...ReaderSampleP
 }
 
 func (p *ReaderSampleProvider) OnBind() error {
+	// If we are not closing on unbind, don't do anything on rebind
+	if p.dontCloseOnUnbind && (p.ivfreader != nil || p.h264reader != nil || p.oggreader != nil) {
+		return nil
+	}
+
 	var err error
 	switch p.Mime {
 	case webrtc.MimeTypeH264:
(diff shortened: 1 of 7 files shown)


boks1971 (approved):

readersampleprovider.go
```diff
@@ -71,6 +73,12 @@ func ReaderTrackWithRTCPHandler(f func(rtcp.Packet)) func(provider *ReaderSample
 	}
 }
 
+func ReaderTrackDisableAutoClose() func(provider *ReaderSampleProvider) {
```
davidzhao, line 76: should it always set this? would be good to ensure we can always correctly republish upon reconnect.

cnderrauber: It requires the application to call close to release resources if it is enabled as we can't release resources in unbind method. Concerning it would cause resource leak in current applications if we enable it by default.

davidzhao: right.. I think the alternatives is that reconnect would produce failures? it seems neither path is ideal.

could Room or Engine close published tracks upon disconnect? that's the path other client SDKs take.

cnderrauber: Thought of this way, but the go-sdk only takes webrtc.TrackLocal interface in `PublishTrack` method that don't have a `Close` method. 
A good way would be to output a error log on republish failure that notify the users to use the new API(new option, and Close method to clean resource). In this way, the application will not have loss after updated the sdk, no resource leak, and no republish improvement but has a message to tell him the right way. Or any other suggestions?

davidzhao: We could have another interface like `TrackLocalWithClose` ? that extends `webrtc.TrackLocal`? and as you suggested:
* if the track isn't a `TrackLocalWithClose`, then log warning so dev remembers to cleanup
* if the track is a `TrackLocalWithClose`, then cleanup automatically.

davidzhao: actually, we might not even require a warning.. since the user is responsible for handling cleaning up in Unbind.. and when republished, Bind will be called again.

cnderrauber: The problem is  if the resource is cleaned in the unbind, bind will failed as the reader is closed.

cnderrauber: > if the track isn't a TrackLocalWithClose, then log warning so dev remembers to cleanup
if the track is a TrackLocalWithClose, then cleanup automatically.

Sounds good, I'll try it

cnderrauber:
The last commit will be an acceptable way, sdk will close tracks that have `Close` method, and `ReadSampleProvider` support republish by default now. If the user uses a custom provider that doesn't support republishing like old `ReadSampleProvider`, he will get error on republishing and is responsible for changing the implementation like `ReadSampleProvider`.

davidzhao (approved):
awesome, lgtm!

room.go
```diff
@@ -194,6 +194,19 @@ func (r *Room) JoinWithToken(url, token string, opts ...ConnectOption) error {
...
+		track := value.(*LocalTrackPublication)
+		if track.Track() != nil {
+			localPubs = append(localPubs, track)
+		}
+		return true
+	})
+
+	for _, pub := range localPubs {
+		pub.CloseTrack()
+	}
```
boks1971, line 209: Can this be moved to a method in `LocalParticipant`? Similar to `republishTracks`. Would be good not to `Range` on an internal element of a different structure from here.

cnderrauber: Done

PR #257, opened 2023-06-14
Fix dropped packet/SN jump handling

frostbyte73:


pkg/jitter/buffer.go
```diff
@@ -192,7 +244,9 @@ func (b *Buffer) pop() []*rtp.Packet {
 	var next *packet
 	for c := b.head; ; c = next {
 		next = c.next
-		packets = append(packets, c.packet)
+		if !c.padding {
```
davidzhao, line 247: 👍

pkg/jitter/buffer.go
```diff
@@ -61,24 +63,45 @@ func (b *Buffer) Push(pkt *rtp.Packet) {
...
 
-	p := b.newPacket(start, end, pkt)
+	p := b.newPacket(start, end, padding, pkt)
+
+	beforePrev := before(pkt.SequenceNumber, b.prevSN)
+	outsidePrevRange := outsideRange(pkt.SequenceNumber, b.prevSN)
+
+	// drop if packet comes before previously pushed packet
+	if b.initialized && beforePrev && !outsidePrevRange {
+		b.logger.Debugw("packet dropped",
```
davidzhao, line 87: if it's a padding packet, I think we shouldn't trigger dropped.. until the next non-padding packet.

davidzhao (approved):

PR #259, opened 2023-06-14
Final jitter updates

frostbyte73:
* Calls onPacketDropped when packets are never received
* Fixes initialization issue where it expected SN to start at 1
* Waits max time after sequence number jump to allow for jitter around the gap
* Fixes issue where it would still push the rest of a sample if the partition head was dropped
* Adds more test scenarios


biglittlebigben (approved):

davidzhao (approved):
lg!

PR #263, opened 2023-06-26
add simulate failure scenario

paulwe:


boks1971 (approved):
I added `SimulateMigrate` and `SimulateServerLeave` also. Just to have more flavours. I will set those up. `NodeFailure` simulation is prevented in `prod` as it kicks out all participants in the room on that node. The other two are participant specific. I can add it after you merge this.

paulwe:
@boks1971 let's merge your version then

boks1971:
Okay, I will pick up your change into my changes.

PR #267, opened 2023-06-28
Fix TrackUnpublished callback being skipped

davidzhao:
- fixed deadlock when triggering Unpublished callback under lock
- removed incorrect comparison causing us to miss unpublished tracks

### DIFF
--- a/remoteparticipant.go
+++ b/remoteparticipant.go
@@ -29,10 +29,6 @@ func (p *RemoteParticipant) updateInfo(pi *livekit.ParticipantInfo) {
 	p.baseParticipant.updateInfo(pi, p)
 	// update tracks
 	validPubs := make(map[string]TrackPublication)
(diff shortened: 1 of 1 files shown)


remoteparticipant.go
```diff
@@ -29,10 +29,6 @@ func (p *RemoteParticipant) updateInfo(pi *livekit.ParticipantInfo) {
 	p.baseParticipant.updateInfo(pi, p)
 	// update tracks
 	validPubs := make(map[string]TrackPublication)
-	p.tracks.Range(func(key, value interface{}) bool {
```
davidzhao, line 32: this was causing us to consider unpublished tracks as valid, breaking unpublished detection mechanism.

boks1971 (approved):

cnderrauber (approved):

Issue #324, opened 2023-10-16
[bug] time.Ticker not close


AtlanCI:
# problem

The ticker that was opened was not closed
https://github.com/livekit/server-sdk-go/blob/main/localsampletrack.go#L427

PR #325, opened 2023-10-16
fix: fix time.Ticker not stop

AtlanCI:
#324


davidzhao (approved):
great catch. Thanks for the PR!

Issue #322, opened 2023-10-09
[bug] goroutine leak (memory leak)


AtlanCI:
# Summary

`RTCEngine.Close` method and `RTCEngine.handleDisconnect.func1` are suspected to have concurrency issues. This can lead to goroutine leaks. The Close code here is really bad, and I don't really understand the logic of Close. 

# Background
In my environment there should be more than 120 users that can connect successfully. After that, there are errors due to insufficient livekit ports reserved. I don't know why the function `RTCEngine.handleDisconnect.func1` is always triggered in my environment. 

# Goroutine leak stack
The stack here is obtained by hacking the package `github.com/pion/interceptor`. 
```golang
goroutine 66030 [running]:
github.com/pion/interceptor/pkg/report.(*ReceiverInterceptor).BindRTCPWriter(0xc00c151500, {0xa19e500?, 0xc04f271db0})
        C:/Users/Atlan/go/pkg/mod/github.com/pion/interceptor@v0.1.17/pkg/report/receiver_interceptor.go:103 +0x10d
github.com/pion/interceptor.(*Chain).BindRTCPWriter(0xc011808960?, {0xa19e500?, 0xc04f271db0?})
        C:/Users/Atlan/go/pkg/mod/github.com/pion/interceptor@v0.1.17/chain.go:30 +0x3f
github.com/pion/webrtc/v3.(*API).NewPeerConnection(0xc0628e2c90, {{0xc05673b540, 0x1, 0x1}, 0x0, 0x0, 0x0, {0x0, 0x0}, {0x0, ...}, ...})
        C:/Users/Atlan/go/pkg/mod/github.com/pion/webrtc/v3@v3.2.16/peerconnection.go:192 +0x758
github.com/livekit/server-sdk-go.NewPCTransport({{0xc05673b540, 0x1, 0x1}, 0x0, 0x0, 0x0, {0x0, 0x0}, {0x0, 0x0, ...}, ...})
        C:/Users/Atlan/go/pkg/mod/github.com/livekit/server-sdk-go@v1.0.15/transport.go:102 +0x618
github.com/livekit/server-sdk-go.(*RTCEngine).configure(0xc0102d7e00, 0xc0178731e0)
        C:/Users/Atlan/go/pkg/mod/github.com/livekit/server-sdk-go@v1.0.15/engine.go:213 +0x285
[...]

PR #331, opened 2023-10-20
fix: goroutine leak

AtlanCI:
#322 

Concurrency exists between user shutdown and automatic reconnection.


cnderrauber:
Thanks for pointing it out, this change still has concurrency issue, I'll create another commit

cnderrauber:
> Thanks for pointing it out! I think this change still has a timing issue, I'll create another commit

I have checked the code again and it doesn't have the timing issue, sorry for the error message. I think it would be better to put all the close code to the `Close` method, so create #332 to fix this.

PR #350, opened 2023-11-18
Close the correct connection.

boks1971:
When trying to resume, server could return a LeaveRequest signalling a full reconnect. In that case, the connection to be closed is the one that was open previously and not the one which received `LeaveRequest`.

Delay caching of conn only if it is an expected initial response (Join OR Reconnect).

Still feels a bit iffy, but addresses the full reconnect in a dozen tries whereas previously it was failing everytime before.


davidzhao (approved):

boks1971:
Closing this in favour of https://github.com/livekit/server-sdk-go/pull/351

PR #351, opened 2023-11-19
Close previous conn when switching to new conn.

boks1971:
This is a different attempt at
https://github.com/livekit/server-sdk-go/pull/350. Feels a bit cleaner, i. e. if the internal `conn atomic.Value` is replaced, the previous `conn` is closed before doing that.

### DIFF
--- a/signalclient.go
+++ b/signalclient.go
@@ -179,7 +179,7 @@ func (c *SignalClient) Close() {
 	if conn != nil {
 		_ = conn.Close()
 	}
-	if isStarted {
+	if isStarted && readerClosedCh != nil {
 		<-readerClosedCh
 	}
 }
--- a/signalclient.go
+++ b/signalclient.go
@@ -181,7 +181,7 @@ func (c *SignalClient) Close() {
 	if isStarted {
 		<-c.readerClosedCh
 	}
-	if conn != nil {
+	if conn != nil && c.OnClose != nil {
 		c.OnClose()
 	}
 }
--- a/signalclient.go
+++ b/signalclient.go
@@ -67,7 +67,7 @@ func (c *SignalClient) Start() {
 		return
 	}
 	c.readerClosedCh = make(chan struct{})
-	go c.readWorker()
+	go c.readWorker(c.readerClosedCh)
 }
 
 func (c *SignalClient) IsStarted() bool {
@@ -174,12 +174,13 @@ func (c *SignalClient) Join(urlPrefix string, token string, params *ConnectParam
 
 func (c *SignalClient) Close() {
 	isStarted := c.IsStarted()
+	readerClosedCh := c.readerClosedCh
 	conn := c.websocketConn()
 	if conn != nil {
 		_ = conn.Close()
 	}
 	if isStarted {
-		<-c.readerClosedCh
+		<-readerClosedCh
 	}
 }
 
@@ -344,11 +345,11 @@ func (c *SignalClient) handleResponse(res *livekit.SignalResponse) {
 	}
 }
 
-func (c *SignalClient) readWorker() {
+func (c *SignalClient) readWorker(readerClosedCh chan struct{}) {
 	defer func() {
 		c.isStarted.Store(false)
 		c.conn.Store(nil)
-		close(c.readerClosedCh)
+		close(readerClosedCh)
 
 		if c.OnClose != nil {
 			c.OnClose()
--- a/signalclient.go
+++ b/signalclient.go
@@ -181,9 +181,6 @@ func (c *SignalClient) Close() {
 	if isStarted {
 		<-c.readerClosedCh
 	}
-	if conn != nil && c.OnClose != nil {
-		c.OnClose()
-	}
 }
 
 func (c *SignalClient) SendICECandidate(candidate webrtc.ICECandidateInit, target livekit.SignalTarget) error {
@@ -352,6 +349,10 @@ func (c *SignalClient) readWorker() {
 		c.isStarted.Store(false)
 		c.conn.Store(nil)
 		close(c.readerClosedCh)
+
+		if c.OnClose != nil {
+			c.OnClose()
+		}
 	}()
 	if pending := c.pendingResponse; pending != nil {
 		c.handleResponse(pending)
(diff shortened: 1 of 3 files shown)


signalclient.go
```diff
@@ -348,18 +350,17 @@ func (c *SignalClient) handleResponse(res *livekit.SignalResponse) {
 func (c *SignalClient) readWorker() {
 	defer func() {
 		c.isStarted.Store(false)
-		if c.OnClose != nil {
-			c.OnClose()
-		}
+		c.conn.Store((*websocket.Conn)(nil))
```
davidzhao, line 353: we cannot store nil into an atomic.Value; it'll panic: https://pkg.go.dev/sync/atomic#Value.Store

would be good to use typed atomic.Pointer instead.

signalclient.go
```diff
@@ -348,18 +350,17 @@ func (c *SignalClient) handleResponse(res *livekit.SignalResponse) {
...
 	}()
 	if pending := c.pendingResponse; pending != nil {
 		c.handleResponse(pending)
 		c.pendingResponse = nil
 	}
-	for !c.isClosed.Load() {
+	for {
 		res, err := c.readResponse()
 		if err != nil {
-			if !isIgnoredWebsocketError(err) && !c.isClosed.Load() {
```
davidzhao, line 362: we'll also need to make sure readResponse returns EOF if it's expected close. otherwise it'll keep spamming logs about error while reading.

signalclient.go
boks1971: Casting `nil` to the concrete type does not cause a panic. Learnt that today. I can't find that post now, but here is a Go Playground example - https://go.dev/play/p/-O0qiTGJxJs>

But, I will change to look at `atomic.Pointer`

signalclient.go
boks1971: Will look at this.

biglittlebigben:
This seems to fix the cloud-ingress restart issue.

signalclient.go
boks1971: On second thought, this  does not do anything different. A unilateral close (by calling Close() method) will close the underlying conn and read will fail with `reading a closed conn`. And this loop will exit/return.

Before this change, the code was only checking `isClosed` which is set by the unilateral `Close()`.

The remote close will be an error returned by `ReadMessage`. Yes, this change will log one message more because it will log an error on an unilateral `Close()` too. But, nothing more.

Once an error is returned, the loop is exited.

Am I missing something?

signalclient.go
biglittlebigben: Right, because a reference to an interface with a type but a nil object is not nil, which makes nil checks really confusing at time. Ask me how I know...

boks1971: How do you know @biglittlebigben 😊?

biglittlebigben: Panic in prod at 3am. It's always a panic in prod at 3am...

signalclient.go
davidzhao: maybe I'm reading this incorrectly.. [here](https://github.com/livekit/server-sdk-go/pull/351/files#diff-db22f9da506d0e6a0874eee29d06d041ca0afa8c9d74681a846af283252f648fL266) it would previously return io.EOF when `Close` has been called due to the `c.isClosed.Load()` check.. currently I think it'll go into `errors.New("cannot read response before join")` ?

It seems that we should not return an error if Close has been called.

davidzhao (approved):

signalclient.go
```diff
@@ -348,18 +350,17 @@ func (c *SignalClient) handleResponse(res *livekit.SignalResponse) {
 func (c *SignalClient) readWorker() {
 	defer func() {
 		c.isStarted.Store(false)
-		if c.OnClose != nil {
-			c.OnClose()
```
cnderrauber, line 352: This means `OnClose` will only be called when `Close` is called instead of connection is interrupted, and the `Engine` relies this callback to handle signal disconnection and resume.

signalclient.go
boks1971: That's correct @davidzhao . You are reading it right. But, what I was saying is that `Close()` in this PR sets the `conn` to `nil` (the other discussion about `atomic.Value == nil`, I will change that to an `atomic.Pointer`). So, yes `readResponse` will return `errors.New("cannot read response before join")`. 

But, the `readLoop` will exit because of that error return.

Only difference is that previously, the read loop would not have logged an error before exiting as it was checking for `isClosed` and with this PR, it will log an error before exiting. So, what I was mentioning was that I could not spot any runaway loops repeating logging an error.

signalclient.go
boks1971: Thank you @cnderrauber . Will take a look. Was hitting a data race in the tests with the `OnClose()` in the `defer`. So, I serialised it :-( My bad. Will take a look at calling back on read loop exit.

signalclient.go
boks1971: Using `atomic.Pointer` in this [commit](https://github.com/livekit/server-sdk-go/pull/351/commits/0e1c8fe0ebcd88598b16d75e0cde2ee29513ddc1).

signalclient.go
boks1971: @cnderrauber Made changes in these two commits
- https://github.com/livekit/server-sdk-go/pull/351/commits/9a727bc78b3482d2a8e4177e8391f80902d19618 - moving `OnClose()` callback to `defer` on `readWorker`.
- https://github.com/livekit/server-sdk-go/pull/351/commits/0acef170f0908065abd1546e9cee319bbeb0ea19 - using a local variable to avoid race. The race reported before is something like [this](https://github.com/livekit/server-sdk-go/actions/runs/6926214815/job/18838032115). The test was calling simulate scenario reconnect which called signal client close. And the read on `readerCloserCh` in `Close()` was being accessed in both `Close()` and `Start()` (where it is changed to set up for next read worker). Getting around it by using a local variable in `Close()` and read on that.

cnderrauber (approved):
Looks good! Thanks for fixing this.

signalclient.go
davidzhao: Sorry, I didn't mean it would be in a loop. Just that in server environments, it would be good to avoid logging errors when the user has closed the connection.

boks1971: Got it, I misunderstood. I will look at adding a change to filter out the error

boks1971: Addressed in this PR - https://github.com/livekit/server-sdk-go/pull/353

PR #357, opened 2023-11-29
Add pacer module for publishing tracks

cnderrauber:
When the go-sdk publishes a high bitrate video track, the instant bitrates might rise too high to overwhelm the network/sfu buffer and cause freeze by packet loss/unorder. This PR adds a pacing module to control the output bitrate at target rate, it will also limit the pending queue latency by increasing the output bitrate temporarily. 
Below code uses the pacer to control output bitrate at 3Mbps and 2s max latency :

```
	pacer := pacer.NewPacerFactory(pacer.LeakyBucketPacer, pacer.WithBitrate(3000000), pacer.WithMaxLatency(2*time.Second))

	room, err := lksdk.ConnectToRoom(host, lksdk.ConnectInfo{
		APIKey:              apiKey,
	}, &lksdk.RoomCallback{
		ParticipantCallback: lksdk.ParticipantCallback{
			OnTrackSubscribed: onTrackSubscribed,
		},
	}, lksdk.WithPacer(pacer))
```

### DIFF
--- a/pkg/interceptor/pacerinteceptor.go
+++ b/pkg/interceptor/pacerinteceptor.go
@@ -1,0 +1,108 @@
+package interceptor
+
+import (
+	"sync"
+
+	"github.com/livekit/mediatransportutil/pkg/pacer"
+	"github.com/pion/interceptor"
+	"github.com/pion/rtp"
+)
+
+type PacerInterceptorFactory struct {
+	pacer pacer.Factory
+	pool  *PacketPool
+}
+
+func NewPacerInterceptorFactory(pacer pacer.Factory) *PacerInterceptorFactory {
+	return &PacerInterceptorFactory{
+		pacer: pacer,
+		pool:  NewPacketPool(500, 1500),
+	}
+}
+
+func (p *PacerInterceptorFactory) NewInterceptor(id string) (interceptor.Interceptor, error) {
+	pacer, err := p.pacer.NewPacer()
+	if err != nil {
+		return nil, err
+	}
+	return &PacerInterceptor{
+		pacer: pacer,
+		pool:  p.pool,
+	}, nil
+}
+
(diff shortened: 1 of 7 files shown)


pkg/interceptor/pacerinteceptor.go
```diff
@@ -0,0 +1,108 @@
...
+		n, err := header.MarshalTo(buf)
+		if err != nil {
+			if pool != nil {
+				pool.Put(poolEntity)
+			}
+			return 0, err
+		}
+		copy(buf[n:], payload)
+		var headCopy rtp.Header
+		headCopy.Unmarshal(buf)
```
boks1971, line 90: Do we need to put together the full packet and unmarshal again to get the header copy? 

In SFU case, the `payload` is in the pool. Here, it is passing along the incoming payload to the pacer, but also creates a pool entity & copies to it. Wondering if that is necessary.

Oh, a copy is needed as pacer could send it later? In that case, I am guessing the `Payload` field 👇 should actuallly point to the `payload` part of the pool entity? Also, wondering if the header can be copied with actually copying it into the pool buffer and unmarshalling it?

boks1971 (approved):
Excellent work as usual @cnderrauber !!!

I could not quite understand the interceptor allocating in pool, copying, unmarshalling and then forwarding the original payload. Left some comments in that area.

pkg/interceptor/pacerinteceptor.go
cnderrauber: > In that case, I am guessing the Payload field 👇 should actuallly point to the payload part of the pool entity
You are right, I was missing it when clean the code, thanks!

> Also, wondering if the header can be copied with actually copying it into the pool buffer and unmarshalling it?
We can't do that because the extensions of the header might point to the underlying buffer that could be reused after Enqueue

PR #364, opened 2023-12-04
Limit rtp packet payload size

cnderrauber:
Limit the rtp packet payload size to 1200 as same as libwebrtc limit to make sure the udp packet don't exceed mtu and udp message payload size in various network & device.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -24,12 +24,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pion/rtp"
 	"github.com/pion/webrtc/v3"
 	"github.com/pion/webrtc/v3/pkg/media"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/atomic"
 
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lg

davidzhao (commented):
Can we also ensure that our packetizer is packetizing to 1200?

davidzhao:
we are performing packetization [here](https://github.com/livekit/server-sdk-go/blob/main/localsampletrack.go#L308)

cnderrauber:
> we are performing packetization [here](https://github.com/livekit/server-sdk-go/blob/main/localsampletrack.go#L308)

@davidzhao The packetizer init here with rtpOuntBoundMTU(1200)
https://github.com/livekit/server-sdk-go/blob/363298fb6d6b1275ef6a353bec8877ec3bc8f4d7/localsampletrack.go#L186

PR #379, opened 2024-01-16
Allow setting participant kind when joining a room

dennwc:
Allow setting participant kind when joining a room. Open source SIP and Cloud server will set it accordingly.


davidzhao (approved):
lg!

PR #385, opened 2024-01-22
Add WriteRTP to LocalSampleTrack API

Sean-Der:
I also renamed `LocalSampleTrack` to `LocalTrack` and add aliases

No API breakage and should be intuitive for users!


davidzhao (approved):
lgtm.

once this merges, mind updating [livekit-cli](https://github.com/search?q=repo%3Alivekit%2Flivekit-cli+NewLocalSampleTrack&type=code) to use the new type?

PR #408, opened 2024-03-02
Support XR request/response to calculate RTT

cnderrauber:


boks1971 (approved):

transport.go
```diff
@@ -130,6 +134,18 @@ func NewPCTransport(params PCTransportParams) (*PCTransport, error) {
...
 	}
 
+	var onXRRtt func(rtt uint32)
+	if params.IsSender {
+		// publisher only responds to XR request for sfu to measure RTT
+		onXRRtt = func(rtt uint32) {}
+	} else {
+		onXRRtt = func(rtt uint32) {
+			t.rttFromXR.Store(true)
+			t.SetRTT(rtt)
```
davidzhao, line 144: Should be calling helper method instead.

```suggestion
			t.setRTT(rtt)
```

PR #411, opened 2024-03-06
Try to keep proper timestamps in WriteSample.

boks1971:
Use

- PacketTimestamp
- Timestamp
- Duration
- PrevDroppedPackets

of `Sample` to figure out RTP timestamp of next sample.

Tested with `livekit-cli` `publishDemo`. Please suggest more testing scenarios.

### DIFF
--- a/localtrack.go
+++ b/localtrack.go
@@ -16,7 +16,9 @@ package lksdk
 
 import (
 	"context"
+	"errors"
 	"io"
+	"math/rand"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
nice!

boks1971:
Going to merge this and tag it. And update `livekit-cli` with it.

biglittlebigben:
Ingress would be affected by this change fairly significantly. It should allow us to remove the code that paces sending the packets to the SDK based on the timestamps, but we need to test this thoroughly in that use case.

PR #412, opened 2024-03-06
lock while accessing engine publisher/subscriber

paulwe:


engine.go
```diff
@@ -388,15 +412,27 @@ func (e *RTCEngine) waitUntilConnected() error {
 }
 
 func (e *RTCEngine) ensurePublisherConnected(ensureDataReady bool) error {
+	e.pclock.Lock()
```
davidzhao, line 415: is it cleaner/simpler to grab the transport under lock? vs having to unlock during waitUntilConnected() ?

engine.go
```diff
@@ -517,11 +553,13 @@ func (e *RTCEngine) resumeConnection() error {
 	e.client.Start()
 
 	// send offer if publisher enabled
+	e.pclock.Lock()
 	if !e.subscriberPrimary || e.hasPublish.Load() {
 		e.publisher.createAndSendOffer(&webrtc.OfferOptions{
```
davidzhao, line 558: also a bit concerned about locking these operations that may perform IO/locks other resources.

davidzhao (commented):
solid fixes, a couple of comments

davidzhao (approved):

PR #418, opened 2024-03-08
Set lastTS to last sample timestamp.

boks1971:
Was using adjusted last time stamp based on number of samples. But, that does not work if somebody is blasting samples. For example, if somebody is blasting 10 seconds worth of samples in 1 second, the adjusted last TS would have moved past the incoming timestamp after the first sample and the subsequent samples would have been dropped as out-of-order.

The change here is to use the incoming time stamp as is for lastTS.

But, this does fail in the scenario of
- Blast 10 seconds worth of audio in one second.
- 20 seconds later, send another sample. In this case, the gap should be 10 seconds, but this change will make it look like 19 seconds.

Need to think more about handling that case. Avoiding the out-of-order strangle for now.

### DIFF
--- a/version.go
+++ b/version.go
@@ -14,4 +14,4 @@
 
 package lksdk
 
-const Version = "2.0.1"
+const Version = "2.0.5"
--- a/localtrack.go
+++ b/localtrack.go
@@ -377,6 +377,9 @@ func (s *LocalTrack) WriteSample(sample media.Sample, opts *SampleWriteOptions) 
 		}
 
 		s.lastTS = sample.Timestamp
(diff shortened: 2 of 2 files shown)


version.go
```diff
@@ -14,4 +14,4 @@
 
 package lksdk
 
-const Version = "2.0.1"
+const Version = "2.0.5"
```
boks1971, line 17: forgot to bump for the last release. So, looks like users are on 2.0.1. Will tag again once this is merged with v2.0.5

davidzhao (approved):

PR #420, opened 2024-03-11
Handle offer errors during resume connection

SijmenHuizenga:
When an error occurs during the webrtc offer creation during connection resume, the resume should be considered failed and an error returned. 

**Why?**

I have seen some situations where the offer creation fails, with the following error message:
```
"msg"="could not negotiate" "error"="ICEAgent does not exist: unable to restart ICETransport"
```
I would have expected to see a retry, but instead the `OnReconnected` is called. This would indicate that the room connection is established, but it is not. The `ParticipantCallback` is never called, and the `OnDataReceived` are quiet. Even when messages are sent to this participant, they are not received.

This change makes sure that errors like this one result in a failure of `resumeConnection()` and thus trigger a proper retry.

To show the impact; the `could not negotiate` happened 51 times in the last 7 days.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -584,9 +584,11 @@ func (e *RTCEngine) resumeConnection() error {
 	publisher := e.publisher
 	e.pclock.Unlock()
 	if sendOffer {
-		publisher.createAndSendOffer(&webrtc.OfferOptions{
+		if err := publisher.createAndSendOffer(&webrtc.OfferOptions{
 			ICERestart: true,
-		})
+		}); err != nil {
+			return err
+		}
 	}
 
 	if err = e.waitUntilConnected(); err != nil {


engine.go
```diff
@@ -584,9 +584,11 @@ func (e *RTCEngine) resumeConnection() error {
 	publisher := e.publisher
 	e.pclock.Unlock()
 	if sendOffer {
-		publisher.createAndSendOffer(&webrtc.OfferOptions{
+		if err := publisher.createAndSendOffer(&webrtc.OfferOptions{
 			ICERestart: true,
-		})
+		}); err != nil {
+			return err
+		}
```
boks1971, line 591: @SijmenHuizenga Thank you for doing this. I think this change is good.

But, hae a few questions
1. Do you have a repro? Wondering if there is something in your set up that makes it easier to repro. Understand this may be an observation in your app and not easily reproduced.
2. There is a `waitForConnection` below this. Does that not error out when the offer creation fails? I would have expected that to fail (although it would have taken the join time out to fail). If that did not fail, looks like there might be more than one issue here.
3. On a retry, does the connection resume and work always? Wondering if we should force a full reconnect on a retry under these conditions.

@davidzhao This change looks good. Have any concerns?

SijmenHuizenga: 1. Unfortunately not. I wish I could reproduce the issue, but I can't. The only thing I have is a pile of logs and user reports.

2. I agree that the `waitForConnection` should detect the failure, but from the logs it seems like it doesn't:

image
(newest logs at the top)

The "Livekit reconnected" is a logline printed inside the `OnReconnected`.  If the `waitForConnection` would fail, i would have expected to see a `resume connection failed` log message, but I don't.

I have the gut feeling that there are more issues causing reconnect problems, but let's tackle them one- at the time :)

3. I'm not able to say for sure if a connection resume always works or always breaks. From a quick glance over the logs of the past two weeks, it seems more often than not a connection resume fails for some reason or another. But this could also be caused by the same reason that the original connection got disconnected (bad network connection).

SijmenHuizenga: We've been running with this change in a small group of our servers. I observe that this change makes re-connect work better:

image

Before it would get stuck in these cases, now it keeps retrying and ultimately fails, triggering an out-of-sdk re-connect.

PR #421, opened 2024-03-12
Revert default behavior change with data publishing

davidzhao:


boks1971 (approved):

dennwc:
I think the expectation about using server SDK is that it abstracts away the packet drop and just delivers app-specific data reliably by default. I think setting it as unreliable is a kind of optimization that app developers may consider (later?). This is why I changed it this way.

Also note that in our own protobuf the default is also `RELIABLE`. So it also matches the generated code better. One less quirk when generating new SDKs.

So I'm wondering why you'd like the default to be `UNRELIABLE`. Is it expected by app developers?

dennwc (approved):

PR #424, opened 2024-03-15
Add OnDisconnectedWithReason room callback

biglittlebigben:
See https://live-kit.slack.com/archives/C02L4JW8ACT/p1710463328128609 for some more context

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -89,8 +89,16 @@ func (cb *ParticipantCallback) Merge(other *ParticipantCallback) {
 	}
 }
 
+type DisconnectionReason string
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
lgtm!

We need to move Go SDK to protocol 13 where `can_reconnect` is obsoleted and server gives the action to perform. I think that can be hidden away in this callback - https://github.com/livekit/protocol/blob/44e9dc34b4ad55b45d87077f7ac8caeb914929b8/protobufs/livekit_rtc.proto#L226

PR #426, opened 2024-03-19
Improve logs related to publishing & unpublishing

SijmenHuizenga:
Added track sid to logs & add unpublish log.

Why? Because it is quite difficult to correlate application-logs and livekit-cloud logs without an Sid.


boks1971 (commented):
Can you please change the key to 'trackID' as we use that in rest of the logging?

boks1971 (approved):

PR #428, opened 2024-03-24
Prevent multi-threaded reading of provider in localtrack

SijmenHuizenga:
tldr; this pr prevents a panic that could happen after a track re-publish.

It seems that it's hypothetically possible for the new `writeWorker` to start running, while the previous `writeWorker` has not yet finished (when `NextSample` is slow). This could cause the `provider.NextSample` to be called twice the same time. 

In the implementation of `NextSample` in `readersampleprovider.go`, we can see that it calls [pion H264Reader.NextNAL()](https://github.com/pion/webrtc/blob/836184c2c1a51ffcc13485f8916b78449414b3d2/pkg/media/h264reader/h264reader.go#L122-L167) which seems not thread-safe at all.

At this time I cannot prove that this situation is causing problems, nor can I reproduce this exact case. It's probably one of those rare race-conditions that almost never happens. But when it would happen, I predict a panic exit.

Tested this change by running it as part of our applications for a bit. Seems to work fine.

### DIFF
--- a/localtrack.go
+++ b/localtrack.go
@@ -64,6 +64,7 @@ type LocalTrack struct {
 	clockRate        float64
 	bound            atomic.Bool
 	lock             sync.RWMutex
+	writeStartupLock sync.Mutex
 	audioLevelID     uint8
 	sdesMidID        uint8
 	sdesRtpStreamID  uint8
@@ -527,18 +528,26 @@ func (s *LocalTrack) rtcpWorker(rtcpReader interceptor.RTCPReader) {
 }
 
 func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
-	s.lock.Lock()
-	if s.cancelWrite != nil {
-		s.cancelWrite()
+	s.writeStartupLock.Lock()
+
+	s.lock.RLock()
+	previousCancel := s.cancelWrite
+	previousWriteClosed := s.writeClosed
+	s.lock.RUnlock()
+
+	if previousCancel != nil {
+		previousCancel()
 		// wait for previous write to finish to prevent multi-threaded provider reading
-		<-s.writeClosed
+		<-previousWriteClosed
 	}
 
+	s.lock.Lock()
 	var ctx context.Context
 	ctx, s.cancelWrite = context.WithCancel(context.Background())
 	writeClosed := make(chan struct{})
 	s.writeClosed = writeClosed
 	s.lock.Unlock()
+	s.writeStartupLock.Unlock()
 	if onComplete != nil {
 		defer onComplete()
 	}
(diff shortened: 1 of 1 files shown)


localtrack.go
```diff
@@ -526,16 +527,22 @@ func (s *LocalTrack) rtcpWorker(rtcpReader interceptor.RTCPReader) {
 }
 
 func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
+	s.lock.Lock()
 	if s.cancelWrite != nil {
 		s.cancelWrite()
+		// wait for previous write to finish to prevent multi-threaded provider reading
+		<-s.writeClosed
```
davidzhao, line 534: Nice fix for locking `s.cancelWrite`. though I'm a bit concerned about holding the lock here while waiting for the previous write to complete.  What do you think about the following instead?

```
s.cancelWrite()
s.cancelWrite = nil
s.lock.Unlock()

<-s.writeClosed
```

SijmenHuizenga: Ah yes, the locking! It's a tricky one for me. I don't see any ways to make this truly thread-safe, without linking the waiting and locking together.

----
This would be your suggestion in context, if I understand correctly?
```
func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
	s.lock.Lock()
	if s.cancelWrite != nil {
		s.cancelWrite()
		s.cancelWrite = nil
		s.lock.Unlock()
		
		<-s.writeClosed		
		s.lock.Lock()
	}
```

Setting `cancelWrite = nil` would cause a third call to `writeWorker` skip waiting for the first call while the second call is waiting for the first one.

---
Another idea:
```
func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
	s.lock.Lock()
	if s.cancelWrite != nil {
		s.cancelWrite()
		s.lock.Unlock()
		<-s.writeClosed
		s.lock.Lock()
	}

```
But then, two calls to `writeWorker` could get stuck waiting for the original worker to finish. Then, when the original writer finishes, one of the waiting writers would get the lock and initialize. But then the next waiting writer would not kill the second writer waiting worker, and would start running parallel to the new worker.

---

The only option to not occupy the `s.lock` during restarts of the writeWorker, would be to introduce a new lock specifically for this function.

Am I missing something? I'm a bit out of idea's on this one :sweat:

SijmenHuizenga: I think I've solved the puzzle: https://github.com/livekit/server-sdk-go/pull/428/commits/b0cfd345a7e8497f26a43deea34fbb8533d26869

davidzhao (approved):
lgtm!

PR #429, opened 2024-03-24
Fix unpublish log name

SijmenHuizenga:
In #426 we added logs to print track unpublications. Now I notice the name seems off:
```
"level"=0 "msg"="unpublished track" "name"="<unhandled-func>" "sid"="TR_VC8SRooGZKzKKb"
```
Let's fix that...

After this change, the log looks like this:
```
"level"=0 "msg"="unpublished track" "name"="front" "sid"="TR_VCBDzfrHRZgNpN"
```


davidzhao (approved):
lg!

PR #434, opened 2024-04-02
Add room parameter to sepcify an interceptor factory to use instead of the defaults. Also allow retrieving the local track SSRC

biglittlebigben:


biglittlebigben:
Still pending a go sdk update

boks1971 (approved):
lgtm!

davidzhao (approved):
lg!

PR #440, opened 2024-04-04
Add WithICETransportPolicy

Sean-Der:
Allows a user to force TURN

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -205,7 +205,10 @@ func (e *RTCEngine) configure(
 	clientConfig *livekit.ClientConfiguration,
 	subscriberPrimary *bool) error {
 	rtcICEServers := FromProtoIceServers(iceServers)
-	configuration := webrtc.Configuration{ICEServers: rtcICEServers}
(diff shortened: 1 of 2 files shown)


examples/filesender/main.go
```diff
@@ -49,7 +50,7 @@ func main() {
 		APISecret:           apiSecret,
 		RoomName:            roomName,
 		ParticipantIdentity: identity,
-	}, &lksdk.RoomCallback{}, lksdk.WithAutoSubscribe(false))
+	}, &lksdk.RoomCallback{}, lksdk.WithAutoSubscribe(false), lksdk.WithICETransportPolicy(webrtc.ICETransportPolicyRelay))
```
davidzhao, line 53: maybe comment out from the example so it doesn't require relay?

davidzhao (approved):

examples/filesender/main.go
Sean-Der: Yes! Oops nice catch 

Neil is doing this for a specific customer, not sure how this will land yet

PR #442, opened 2024-04-10
Use slog in a backward compatible way

dennwc:
Go SDK will be used in other projects, thus it makes sense to use standard structured logging instead of our own one.

https://github.com/livekit/protocol/pull/668 added support for `slog` to the protocol's `logger` package. This change propagates it to the SDK.

For our own projects, calling `logger.SetLogger` will automatically set `slog` logger as well, making this change backward-compatible.

Also, add `SetLogger` to the most important structs of the package, allowing to associate additional metadata with RTC logs.


davidzhao (commented):
Did we have to change all of the logger usage internally to slog?

IMO it makes sense to use it as the interface, and when interfacing with external clients.  But using slog internally in the code base seems to be a step backwards for the following reasons:
* It breaks a certain consistency within our Go codebase. Everywhere else we are using protocol logger
* We would lose features, particularly with [WithValues](https://github.com/livekit/protocol/blob/main/logger/logger.go#L90), WithComponent, etc

dennwc:
> Did we have to change all of the logger usage internally to slog?

Only in the SDK, yes.

> It breaks a certain consistency within our Go codebase. Everywhere else we are using protocol logger

This depends on if we will adopt it long term. It could be just a temporary inconsistency.

In practice (if I understand correctly), we mostly use Zap under the hood, which has slog adapter as well. So in general we could decide to move to slog as a main log interface eventually. That would remove the need to maintain our own logger interface. There are too many loggers in the wild already :sweat_smile:

But that's just as an option, it's not directly related to the PR.

> We would lose features, particularly with [WithValues](https://github.com/livekit/protocol/blob/main/logger/logger.go#L90), WithComponent, etc

`WithValues` is supported: [slog.Logger.With](https://pkg.go.dev/log/slog#Logger.With). `WithComponent` could be a wrapper for `With("some-predefined-key", name+component)` in our `logger` package. I will make sure it plays nicely with Zap if we decide to switch.

We do lose Zap-specific things like `WithItemSampler`. But I think it could also be a wrapper in `logger` package that asserts that the slog handler is Zap and reconfigures it accordingly. Something along the lines of:

```go
package logger

func WithItemSampler(log *slog.Logger) *slog.Logger
```

If the underlying logger is not Zap it will just return the same logger with no changes. This will work for Logr, for example, but also for any other log package out there.

So it's definitely worth a separate discussion about the long term plans. I could make a separate DD/proposal with all the details that we can discuss.

But for now this PR only changes client-facing SDK to use the logger from Go stdlib. I think eventually most libraries will converge to using that interface anyway, so I wanted to see how invasive it would be to switch. Turns out it's not.

PR #443, opened 2024-04-11
Properly update SID mapping on participant update

dennwc:
During the work on E2E tests for SIP I noticed that SDK misbehaves if one attempts to re-create a participant with the same identity after leaving.

SDK updates the participant's SID, but doesn't update the SID->Identity mapping. This leads to media tracks dropped because they do not match any participant SIDs.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -29,6 +29,8 @@ import (
 
 var Default = Build
 
+const livekitServerVersion = "v1.5" // TODO: switch to latest when the issue is resolved
+
 func Build() error {
 	fmt.Println("building...")
 	cmd := exec.Command("go", "build", ".")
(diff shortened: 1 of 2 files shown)


dennwc:
Can't reproduce CI failure locally :thinking:

biglittlebigben (approved):
LGTM. The test failure is something to look into indeed though.

dennwc:
After some trial and error, I restarted the tests on the current `main` branch of SDK and they [fail the same way](https://github.com/livekit/server-sdk-go/actions/runs/8574335111/job/23837899970) as this PR.

Tests pass on Docker image `livekit/livekit-server@v1.5.3` and fail on `livekit/livekit-server:latest` (which is `v1.6.0`).

I'll downgrade LK version used in tests for now and will merge this PR if CI is green.

PR #455, opened 2024-04-26
Created echo example project

s-hamdananwar:
When I was trying to use the Go SDK for implementing audio subscription and publication requirements, I couldn't find any examples showcasing how to work with audio packet payloads that can be then used for different purposes like transcoding into different codecs, sending/receiving as multicast packets etc. 
I got great help from @Sean-Der through Slack on how to implement some of my requirements, and my goal with this simple example project is to show a very basic implementation of how to access audio packets from a track, as well as how to write audio packets into a track. I believe this example can be used as starter code for anyone with audio processing requirements that deals with subscribing tracks from and/or publishing tracks into LiveKit.

### DIFF
--- a/examples/echo/main.go
+++ b/examples/echo/main.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/pion/webrtc/v3"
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm! thanks for the contribution 🎉

ATtendev:
@s-hamdananwar Could you add a video track example? 
Thanks.

PR #482, opened 2024-06-09
Expose Room.ConnectionState()

davidzhao:


room.go
```diff
@@ -152,6 +163,7 @@ func NewRoom(callback *RoomCallback) *Room {
 		sidDefers:          make(map[livekit.ParticipantID][]func(*RemoteParticipant)),
 		callback:           NewRoomCallback(),
 		sidReady:           make(chan struct{}),
+		connectionState:    ConnectionStateDisconnected,
```
boks1971, line 166: Maybe a 'ConnectionStateInit' as disconnected is terminal and it would be good to have one meaning for disconnected value.

boks1971 (approved):

paulwe (approved):

room.go
davidzhao: it's not a bad idea.. though if we were to add the state we should do so across clients. Currently the other client SDKs do not expose this `initial` state: https://github.com/livekit/client-sdk-js/blob/main/src/room/Room.ts#L104

davidzhao: so I'll merge this one without. we can add in the future if we decide to change.

PR #489, opened 2024-06-20
Use SIP grants

dennwc:


auth.go
```diff
@@ -27,9 +27,12 @@ type authBase struct {
 	apiSecret string
 }
 
-func (b authBase) withAuth(ctx context.Context, grant auth.VideoGrant) (context.Context, error) {
+func (b authBase) withAuth(ctx context.Context, grant auth.VideoGrant, sip *auth.SIPGrant) (context.Context, error) {
```
davidzhao, line 30: I think it'd be better to have two separate functions for these things, since we very rarely use these together

davidzhao: i.e. `withSIPAuth`

dennwc: Updated, kept a single method, but used interface type as argument. Looks a lot nicer. WDYT?

Should probably be moved to `protocol/auth` eventually.

davidzhao (approved):
nice, even better!

PR #493, opened 2024-06-27
Support setting attributes

dennwc:


biglittlebigben (approved):

participant.go
```diff
@@ -201,13 +210,29 @@ func (p *baseParticipant) updateInfo(pi *livekit.ParticipantInfo, participant Pa
 	p.name = pi.Name
 	oldMetadata := p.metadata
 	p.metadata = pi.Metadata
-	hasChanged := oldMetadata != p.metadata
+	metaChanged := oldMetadata != p.metadata
+	attrsChanged := len(pi.Attributes) != 0
```
davidzhao, line 214: this is incorrect. the server sends down all attributes with each update, given they are idempotent. what should be done here is to compare the attributes with existing values, and firing the event if any of the values in `pi.Attributes` do not match `p.attributes`

dennwc: Hmm, I see. Will fix it :+1:

PR #494, opened 2024-06-27
Bake outbound def timeout from CLI into SDK

dennwc:
Bake outbound default timeout [from CLI](https://github.com/livekit/livekit-cli/pull/340#discussion_r1653484970) into SDK. Our default timeout is too short for this specific API.

### DIFF
--- a/sipclient.go
+++ b/sipclient.go
@@ -17,6 +17,7 @@ package lksdk
 import (
 	"context"
 	"net/http"
+	"time"
 
 	"github.com/livekit/protocol/livekit"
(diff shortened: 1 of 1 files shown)


frostbyte73 (approved):

frostbyte73:
Isn't this only adding a 30s timeout if there isn't already a timeout?

dennwc:
@frostbyte73 that's correct, but the client lib then does a similar check under the hood and sets an even lower timeout (5 sec?). So without setting a higher one here all `CreateSIPParticipant` requests will timeout.

PR #496, opened 2024-06-28
Fix attribute update

dennwc:
Previous attribute update assumed server update sends delta attributes and not the whole state. This PR fixes this assumption and overrides attributes with ones sent by the server.

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -20,12 +20,17 @@ import (
 	"github.com/livekit/protocol/livekit"
 )
 
+// ParticipantAttributesChangedFunc is callback for Participant attribute change event.
+// The function is called with an already updated participant state and the map of changes attributes.
+// Deleted attributes will have empty string value in the changed map.
+type ParticipantAttributesChangedFunc func(changed map[string]string, p Participant)
+
 type ParticipantCallback struct {
 	// for all participants
 	OnTrackMuted               func(pub TrackPublication, p Participant)
 	OnTrackUnmuted             func(pub TrackPublication, p Participant)
 	OnMetadataChanged          func(oldMetadata string, p Participant)
-	OnAttributesChanged        func(oldAttrs map[string]string, p Participant)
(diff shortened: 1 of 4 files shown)


frostbyte73 (approved):

davidzhao (commented):
lg. one thing I'm wondering is the ergonomics of this interface. it seems more useful to send an update of only the things that have changed? wdyt?

cc: @lukasIO for feedback as well

lukasIO:
> I'm wondering is the ergonomics of this interface 

I'm not sure I understand which interface you are referring to

> it seems more useful to send an update of only the things that have changed

in case this is meant from an API perspective, I think ultimately the most desirable for users would be to be able to pass a specific attribute (or list of attributes) that they're interested in and be able to only be notified for changes of that attribute.
From a more general perspective, I think that both the full state and changed-only state could be useful to users, but the full state is more important, I think.

dennwc:
To recap the Slack discussion, we will pass the diff to the callbacks.

The only question is what will indicate a delete in the map on both the setter and callback side.

We think that passing empty string across all SDK will make most sense. Otherwise some SDKs would accept `null` and will error on empty string (because that's not a valid attribute value), and some won't. Allowing both `null` and empty string is also confusing, since empty attributes won't be preserved.

Later we can add a separate API to subscribe to specific attribute changes with explicit "delete" type, e.g.:
```go
const (
   Added = Change(iota)
   Changed
   Deleted
)

func OnAttributeChange(keys []string, callback func(key, val string, change Change))
```

lukasIO (approved):

biglittlebigben:
Is there any scenario where a client may miss an update, and if so, how are they expected to eventually recover the proper state?

dennwc:
@biglittlebigben luckily, server always sends the full participant info snapshot. Attribute delta is computed only for the client.

I initially assumed server sends delta, but this is incorrect. So this PR fixes SDK, which now assumes that the full snapshot is sent.

PR #499, opened 2024-07-09
Fix receiving data messages for local participant

dennwc:
Currently the data callback is called on the message sender (and the room) instead of a local participant.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -485,11 +485,11 @@ func (e *RTCEngine) handleDataPacket(msg webrtc.DataChannelMessage) {
 		if onDataReceived := e.OnDataReceived; onDataReceived != nil {
 			onDataReceived(m)
 		}
+		if identity == "" {
+			//lint:ignore SA1019 backward compatibility
+			identity = m.ParticipantIdentity
+		}
 		if e.OnDataPacket != nil {
-			if identity == "" {
-				//lint:ignore SA1019 backward compatibility
-				identity = m.ParticipantIdentity
-			}
 			e.OnDataPacket(identity, &UserDataPacket{
 				Payload: m.Payload,
 				Topic:   m.GetTopic(),
--- a/room.go
+++ b/room.go
@@ -457,26 +457,28 @@ func (r *Room) handleResumed() {
 	r.sendSyncState()
 }
 
-func (r *Room) handleDataReceived(identity string, dataPacket DataPacket) {
-	if identity == r.LocalParticipant.Identity() {
+func (r *Room) handleDataReceived(senderIdentity string, dataPacket DataPacket) {
+	if senderIdentity == r.LocalParticipant.Identity() {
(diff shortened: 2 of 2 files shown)


engine.go
```diff
@@ -485,11 +485,11 @@ func (e *RTCEngine) handleDataPacket(msg webrtc.DataChannelMessage) {
 		if onDataReceived := e.OnDataReceived; onDataReceived != nil {
 			onDataReceived(m)
 		}
+		if identity == "" {
+			//lint:ignore SA1019 backward compatibility
+			identity = m.ParticipantIdentity
```
biglittlebigben, line 490: Is this not needed anymore?

room.go
```diff
@@ -457,26 +457,28 @@ func (r *Room) handleResumed() {
...
 		return
 	}
-	p := r.GetParticipantByIdentity(identity)
 	params := DataReceiveParams{
-		SenderIdentity: identity,
-		Sender:         p,
+		SenderIdentity: senderIdentity,
+		Sender:         r.GetParticipantByIdentity(senderIdentity),
 	}
+	// only local participant should receive data callbacks
```
biglittlebigben, line 469: This seems like an API change. Is there any reason? Has it been discussed?

biglittlebigben (approved):

engine.go
dennwc: It's just moved out of the condition

room.go
dennwc: I may be misunderstanding it, but when user connect with SDK he can set two data callback: on the room and on the (local) participant. And the second callback doesn't fire when data message is received. I'm not sure why it was set to fire on a _remote_ participant instead of the local. Looks like a bug to me.

biglittlebigben: Is it meant to fire on the participant object representing the metadata sender or the receiver?

davidzhao: it's meant to be fired on the sender. docs here: https://github.com/livekit/client-sdk-js/blob/main/src/room/events.ts#L424

dennwc: That's pretty confusing, I must admit.

Why do we have `OnDataPacket` on the `LocalParticipant` then? If the logic is "fire at the sender", local participant [will always skip](https://github.com/livekit/server-sdk-go/pull/499/commits/ed06c90dca2971b504ec44bff439772589dab4a5#diff-49f4dcd9caec275f255f290d41ae11cc0e47384c4f45c9ff2256bf80ec639e12R461) their own messages.

biglittlebigben: I'm not sure, but reach out more widely to folks who have been maintaining that piece of code to understand the intended behavior.

davidzhao: yeah.. makes no sense on the `LocalParticipant`. There are two ways of receiving data:
1. give me all data messages (listen on room)
2. give me all messages sent from a specific remote participant (listen on RemoteParticipant)

PR #513, opened 2024-08-06
use livekit cloud region settings endpoint to find best edge datacenter

matkam:


regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+			Timeout: 5 * time.Second,
+		},
+	}
+}
+
+func (r *regionURLProvider) RefreshRegionSettings(cloudHostname, token string) error {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+
+	return r.refreshRegionSettings(cloudHostname, token)
```
boks1971, line 48: Would be good to avoid this under lock as this is making a network request.

regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// BestURL returns the region with least number of attempts, in the order provided by the region settings endpoint (round-robin)
+func (r *regionURLProvider) BestURL(cloudHostname, token string) (string, error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+
+	hostnameSettings := r.hostnameSettingsCache[cloudHostname]
+	if hostnameSettings == nil || time.Since(hostnameSettings.updatedAt) > regionHostnameProviderSettingsCacheTime {
+		if err := r.refreshRegionSettings(cloudHostname, token); err != nil {
```
boks1971, line 58: here as well, it is under lock.

regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+
+type hostnameSettingsCacheItem struct {
+	regionSettings    *livekit.RegionSettings
+	updatedAt         time.Time
+	regionURLAttempts map[string]int
+}
+
+func newRegionURLProvider() *regionURLProvider {
+	return &regionURLProvider{
+		hostnameSettingsCache: make(map[string]*hostnameSettingsCacheItem),
```
boks1971, line 37: any specific reason to support multiple cloud host names? Guessing this is going to be used for a project.

boks1971 (approved):
Nice!! Have some questions inline.

regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if !isCloud(parsedURL.Hostname()) {
+		return "", errors.New("not a cloud url")
+	}
+
+	return parsedURL.Hostname(), nil
+}
+
+func isCloud(hostname string) bool {
+	return strings.HasSuffix(hostname, "livekit.cloud") || strings.HasSuffix(hostname, "livekit.io")
```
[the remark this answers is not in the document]
boks1971: Would be good to make this a config so that we can add/remove domains easily without changing code. Can also address the private cloud case. If config is empty, can return `false`.

regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+	// remove failed region from regionSettings
+	for i, region := range hostnameSettings.regionSettings.Regions {
+		if region.Url == regionURL {
+			hostnameSettings.regionSettings.Regions = append(
+				hostnameSettings.regionSettings.Regions[:i],
+				hostnameSettings.regionSettings.Regions[i+1:]...,
+			)
+			break
+		}
+	}
```
paulwe, line 100: `slices.DeleteFunc`?

regionurlprovider.go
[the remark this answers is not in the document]
paulwe: could we add the regions endpoint to oss so the request returns a url from config or formatted from the http host header? pattern based detection in the client feels fragile. if we want oss clients to return an error the endpoint could send back a non 200 response with a meaningful error message.

paulwe (approved):

regionurlprovider.go
matkam: no specific reason. I just didn't know how people were going to be using this SDK and wanted to keep it flexible.

regionurlprovider.go
```diff
@@ -89,54 +129,9 @@ func (r *regionURLProvider) ReportAttemptFailure(cloudHostname, regionURL string
...
-	if err := protojson.Unmarshal(respBody, regions); err != nil {
-		return errors.New("refreshRegionSettings failed to decode region settings: " + err.Error())
-	}
-
-	r.hostnameSettingsCache[cloudHostname] = &hostnameSettingsCacheItem{
-		regionSettings:    regions,
-		updatedAt:         time.Now(),
-		regionURLAttempts: map[string]int{},
-	}
+	_ = slices.DeleteFunc(hostnameSettings.regionSettings.Regions, func(region *livekit.RegionInfo) bool {
```
paulwe, line 132: the result needs to be assigned back to `hostnameSettings.regionSettings.Regions` - it works like `append`

regionurlprovider.go
```diff
@@ -0,0 +1,159 @@
...
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return errors.New("refreshRegionSettings failed to read response body: " + err.Error())
+	}
+	regions := &livekit.RegionSettings{}
+	if err := protojson.Unmarshal(respBody, regions); err != nil {
+		return errors.New("refreshRegionSettings failed to decode region settings: " + err.Error())
+	}
+
```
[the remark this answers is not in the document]
matkam: lmk if the current change works for you, or if you think I should use locking around an anonymous function instead

room.go
```diff
@@ -266,9 +288,31 @@ func (r *Room) JoinWithToken(url, token string, opts ...ConnectOption) error {
...
+		var err error
+		joinRes, err = r.engine.Join(url, token, params)
+		if err != nil {
+			return err
+		}
+	} else {
+		for joinRes == nil {
+			bestURL, err := r.regionURLProvider.BestURL(cloudHostname, token)
+			if err != nil {
+				return err
```
real-danm, line 303: if for whatever reason we can't find the best url here, should this just log an error instead and let it continue which will just try to join using the latency record?

regionurlprovider.go
```diff
@@ -0,0 +1,156 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/livekit/protocol/livekit"
+)
+
+const (
+	regionHostnameProviderSettingsCacheTime = 3 * time.Second
+)
+
+type regionURLProvider struct {
```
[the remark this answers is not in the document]
matkam: we've been using region/cluster pretty interchangeably. I'm not sure what the right answer is here.

regionurlprovider.go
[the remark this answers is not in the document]
matkam: I like the idea of avoiding code path changes based on string pattern matching. But I'll leave that for a subsequent PR as the JS SDK currently does this, and we still have more to discuss on this topic.

room.go
```diff
@@ -229,6 +241,16 @@ func (r *Room) SID() string {
 	return r.sid
 }
 
+// PrepareConnection - with LiveKit Cloud, determine the best edge data center for the current client to connect to
+func (r *Room) PrepareConnection(url, token string) error {
```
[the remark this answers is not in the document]
matkam: I added this usage to `examples/echo` and made a reference to the `examples` dir in the README

room.go
```diff
@@ -266,9 +288,31 @@ func (r *Room) JoinWithToken(url, token string, opts ...ConnectOption) error {
...
+			}
+
+			logger.Debugw("RTC engine joining room",
+				"url", bestURL,
+			)
+			joinRes, err = r.engine.Join(bestURL, token, params)
+			if err != nil {
+				// try the next URL
+				r.regionURLProvider.ReportAttemptFailure(cloudHostname, bestURL)
+				continue
```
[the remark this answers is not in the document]
matkam: I think we do want to retry with the next best cluster/region without a delay. the JS SDK looks for either a 401 or a timeout before a retry, instead of always retrying. I wonder if that makes more sense, like this:
```
if err != nil {
  if os.IsTimeout(err) || strings.HasPrefix(err.Error(), "unauthorized: ") {
    // try the next URL
    r.regionURLProvider.ReportAttemptFailure(cloudHostname, bestURL)
    continue
  }
  return err
}
```

regionurlprovider.go
```diff
@@ -0,0 +1,156 @@
...
+	return nil
+}
+
+// BestURL returns the region with least number of attempts, in the order provided by the region settings endpoint (round-robin)
+func (r *regionURLProvider) BestURL(cloudHostname, token string) (string, error) {
+	r.mutex.RLock()
+	hostnameSettings := r.hostnameSettingsCache[cloudHostname]
+	r.mutex.RUnlock()
+
+	if hostnameSettings == nil || time.Since(hostnameSettings.updatedAt) > regionHostnameProviderSettingsCacheTime {
```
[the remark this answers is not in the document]
matkam: in the tight loop scenario, I think pounding `RefreshRegionSettings()` is better than not having URLs available in subsequent calls.

room.go
matkam: I'm not sure about this one. In what scenario would the region settings endpoint not be available, but the websocket/signal connection would still go through? This would just add another step before eventual failure.

matkam: changed the logic around to always fallback to the original url on any failure

PR #537, opened 2024-10-02
Allow setting disconnect reason. Map SIP reasons.

dennwc:
Allow setting disconnect reason. Map new SIP disconnect reasons to SDK constants.

Requires https://github.com/livekit/protocol/pull/828


engine.go
```diff
@@ -661,12 +662,13 @@ func (e *RTCEngine) handleLeave(leave *livekit.LeaveRequest) {
...
+		reason := leave.GetReason()
 		e.log.Infow("server initiated leave",
-			"reason", leave.GetReason(),
+			"reason", reason,
 			"canReconnect", leave.GetCanReconnect(),
 		)
 		if e.OnDisconnected != nil {
-			e.OnDisconnected(LeaveRequested)
+			e.OnDisconnected(GetDisconnectionReason(reason))
 		}
```
boks1971, line 672: not about this PR, but we need to update this SDK to support LeaveRequest.Action at some point

callback.go
```diff
@@ -107,10 +107,24 @@ func (cb *ParticipantCallback) Merge(other *ParticipantCallback) {
 type DisconnectionReason string
 
 const (
-	LeaveRequested DisconnectionReason = "leave requested by room"
-	Failed         DisconnectionReason = "connection to room failed"
+	LeaveRequested  DisconnectionReason = "leave requested by room"
+	UserUnavailable DisconnectionReason = "remote user unavailable"
+	RejectedByUser  DisconnectionReason = "rejected by remote user"
+	Failed          DisconnectionReason = "connection to room failed"
```
boks1971, line 113: Would be nice to move protocol `DisconnectReason` -> `string` to common place like `protocol` and use it both here and also in `livekit` server. Comment not related to this PR.

boks1971 (approved):

engine.go
dennwc: Will add a TODO here :+1:

callback.go
dennwc: Actually this might be the right time. Since I need to update both anyway, I could do it as well.

The only problem is that `OnDisconnected` callback accepts this string type currently and changing it to something else would not be backward-compatible. Will have to add a new callback, probably.

boks1971: Do you know if somebody is telling on that string value? Guess it is impossible to say.

We can keep the same string for the two existing there and add the rest.

boks1971: Oh string type. Got it. Yeah that's messy

PR #548, opened 2024-10-23
Fast publish

cnderrauber:
* Establish publisher connection when joining room if possible
* Parallel negotiate webrtc track with signal request
* Reduce negotiate interval


boks1971 (approved):
🚀

Issue #555, opened 2024-11-12
Track ids get messed up when publishing two tracks at the same time


SijmenHuizenga:
At about the exact same moment, two tracks were being published in the same room. The track ID's got messed up, it seems inside the sdk. Running sdk version 15f576aa52c50ee6eff4b63ceb1c9d9b7956f4c1.  Race condition? 

The following log lines happen practically at the same time:

1. The livekit SDK logs that track with name `front-xs#kaTUEk` is published with track ID `TR_VChsPfQLKR5vv8`:
```
2024/11/12 06:03:11 INF [livekit-sdk] ts=2024-11-12 06:03:11.339546 caller={file=logger.go line=440} level=0 msg=published track name=front-xs#kaTUEk source=CAMERA trackID=TR_VChsPfQLKR5vv8
```

Then `OnLocalTrackPublished` is called with track name `front-xs#kaTUEk` and track ID `TR_VCAy25euEr8rdM` ❌. This is the wrong track id for this track name. As you can see in the full logs below the track Id `TR_VCAy25euEr8rdM` belongs to a track named `left#sZhnyh`.
```
2024/11/12 06:03:11 INF [app] [room robot-zora-10 0] [RM_AipbuMjjhbph] local track published front-xs#kaTUEk TR_VCAy25euEr8rdM
```
This log line originates from our app:
```
livekit.GetLocalParticipant().Callback.OnLocalTrackPublished = func(publication *lksdk.LocalTrackPublication, lp *lksdk.LocalParticipant) {
    sid := publication.SID()
    livekit.Log(logger.Info, "local track published %s %s", publication.Name(), sid)
```

full logs:

```
2024/11/12 06:03:11 INF [app] [room robot-zora-10 0] [RM_AipbuMjjhbph] local track published left#sZhnyh TR_VChsPfQLKR5vv8
2024/11/12 06:03:11 INF [app] [room robot-zora-10 0] [RM_AipbuMjjhbph] [left#sZhnyh] [] SID changed from  to TR_VChsPfQLKR5vv8
2024/11/12 06:03:11 INF [livekit-sdk] ts=2024-11-12 06:03:11.339546 caller={file=logger.go line=440} level=0 msg=published track name=front-xs#kaTUEk source=CAMERA trackID=TR_VChsPfQLKR5vv8
[...]

PR #556, opened 2024-11-12
Address publish race.

boks1971:
With a single channel to receive track published response, responses could have been mixed up with requests. Demux using request Cid.

For https://github.com/livekit/server-sdk-go/issues/555

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -37,17 +37,22 @@ const (
 )
 
 type RTCEngine struct {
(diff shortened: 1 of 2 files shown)


SijmenHuizenga (approved):
Nice fix! I can totally see how this shared-channel could have caused #555. We tested this branch with a room and quite some tracks and works good :+1:

PR #573, opened 2024-12-09
Add audio encryption/decryption utilities

skheyfets-asapp:
Added:
- Key derivation functions (DeriveKeyFrom*) from string and byte array which derive the same keys as `client-sdk-js` with the same inputs
- Encryption/Decryption utilities for audio packets which encrypt/decrypt using AES-GCM 128bit and compatible with `client-sdk-js`
- Decryption utility will also drop the server injected frames

Changed:

- TrackPublicationOptions struct to support passing the Encryption Type (already part of protocol)
- LocalParticipant.PublishTrack() to pass the encryption type from TrackPublicationOptions into signaling
- Room to store `sifTrailer` from joinResponse so that `sifTrailer` can be passed later to Decryption utility


encryption.go
```diff
@@ -0,0 +1,204 @@
...
+func DecryptGCMAudioSample(sample, key, sifTrailer []byte) ([]byte, error) {
+
+	if len(key) != 16 {
+		return nil, ErrIncorrectKeyLength
+	}
+
+	if sifTrailer != nil && len(sample) >= len(sifTrailer) {
+		possibleTrailer := sample[len(sample)-len(sifTrailer):]
+		if bytes.Equal(possibleTrailer, sifTrailer) {
+			// this is unencrypted Server Injected Frame (SIF) thas should be dropped
```
lukasIO, line 98:
```suggestion
			// this is unencrypted Server Injected Frame (SIF) that should be dropped
```

lukasIO (approved):
nice, this lgtm!

encryption_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func TestDeriveKeyFromBytes(t *testing.T) {
+
+	inputSecret := []byte{34, 21, 187, 202, 134, 204, 168, 62, 5, 105, 40, 244, 88}
+	expectedKey := []byte{129, 224, 93, 62, 17, 203, 99, 136, 101, 35, 149, 128, 189, 152, 251, 76}
+
+	key, err := DeriveKeyFromBytes(inputSecret)
+	assert.Nil(t, err)
+	assert.Equal(t, expectedKey, key)
```
boks1971, line 42: Can we switch these to `require` package? LK has been using it in other repos.

encryption.go
```diff
@@ -0,0 +1,204 @@
...
+	iv := make([]byte, ivLength)
+	copy(iv, sample[ivStart:ivStart+ivLength]) // copy IV value out of sample into iv
+
+	cipherTextStart := len(frameHeader)
+	cipherTextLength := len(sample) - len(frameTrailer) - ivLength - len(frameHeader)
+	cipherText := make([]byte, cipherTextLength)
+	copy(cipherText, sample[cipherTextStart:cipherTextStart+cipherTextLength])
+
+	// setup AES
+	aesCipher, err := aes.NewCipher(key)
```
boks1971, line 124: Is there any performance penalty to creating this for every sample? Can this be created in track context and re-used?

boks1971 (commented):
lgtm, some question inline. Sorry for the late review.

encryption_test.go
skheyfets-asapp: sounds good, I'll have a look

encryption.go
skheyfets-asapp: I've tried to keep it as separate and self-contained as possible without the need to instantiate anything separately to use it. I believe the AES operations are highly optimized in Go with the use of Go Assembler, but I'll see if I can run a quick benchmark with and without the cipher reuse.

PR #575, opened 2024-12-09
Support get for SIP resources.

dennwc:


sipclient.go
```diff
@@ -68,6 +68,50 @@ func (s *SIPClient) CreateSIPOutboundTrunk(ctx context.Context, in *livekit.Crea
...
+	if err != nil {
+		return nil, err
+	}
+	// Client-side filtering, in case SDK is newer than the server.
+	return req.FilterSlice(resp.Items), nil
+}
+
+// GetSIPOutboundTrunksByID gets SIP Outbound Trunks by ID.
+// Returned slice is in the same order as the IDs. Missing IDs will have nil in the corresponding position.
+func (s *SIPClient) GetSIPOutboundTrunksByID(ctx context.Context, ids []string) ([]*livekit.SIPOutboundTrunkInfo, error) {
```
davidzhao, line 95: nit, should these be
```suggestion
func (s *SIPClient) GetSIPOutboundTrunksByID(ctx context.Context, ids ...string) ([]*livekit.SIPOutboundTrunkInfo, error) {
```

or renaming the function to refer to plural? `GetSIPOutboundTrunksByIDs`?

dennwc: An alternative could also be:
```
GetSIPOutboundTrunks(ctx context.Context, ids ...string)
```
And remove the requirement for non-empty IDs list. If no IDs are listed - return all the trunks.

davidzhao (approved):

PR #583, opened 2024-12-17
Add E2EE example, style changes, more tests

skheyfets-asapp:
- Addressed code style comments
- Switched to `require` testifiy package in encryption tests
- Refactored and added encryption utilities with ability to reuse AES cipher
- Added echo example with E2EE support (audio decrypted and reencrypted before sending back)
- Added benchmark tests for encryption utilities (results from dev machine below)
```
goos: darwin
goarch: amd64
pkg: github.com/livekit/server-sdk-go/v2
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkDecryptAudioNewCipher-12                1781293               603.1 ns/op          1104 B/op          6 allocs/op
BenchmarkDecryptAudioCachedCipher-12             2778477               429.1 ns/op           592 B/op          5 allocs/op
BenchmarkEncryptAudioCachedCipher-12             1287099               884.3 ns/op           632 B/op          6 allocs/op
BenchmarkEncryptAudioNewCipher-12                1080866              1035 ns/op            1144 B/op          7 allocs/op
```

### DIFF
--- a/encryption.go
+++ b/encryption.go
@@ -22,11 +22,14 @@ const (
 	unencrypted_audio_bytes  = 1
 )
 
-var ErrIncorrectKeyLength = errors.New("incorrect key length for encryption/decryption")
-var ErrUnableGenerateIV = errors.New("unable to generate iv for encryption")
-var ErrIncorrectIVLength = errors.New("incorrect iv length")
-var ErrIncorrectSecretLength = errors.New("input secret provided to derivation function cannot be empty or nil")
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
Thank you. Guess not a huge performance win, but a sizeable one!

boks1971 (commented):
Thank you for doing this @skheyfets-asapp . I will merge this now.

PR #588, opened 2024-12-27
version bump to 2.4.1

davidzhao:


paulwe (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/livekit/server-sdk-go/v2
 
-go 1.23
+go 1.22.7
```
paulwe, line 3: was this intentional?

davidzhao: yes, it was raised unintentionally in protocol and was never rolled back in the SDK side

PR #595, opened 2025-01-16
Do not fail to initialize the jitter buffer if the first enqueued buffer is not a partition start

biglittlebigben:
This drops buffers that are sent to the jitter buffer before the first partition start

### DIFF
--- a/pkg/jitter/buffer.go
+++ b/pkg/jitter/buffer.go
@@ -15,7 +15,6 @@
 package jitter
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cnderrauber (approved):

pkg/jitter/buffer.go
```diff
@@ -105,6 +104,9 @@ func (b *Buffer) Push(pkt *rtp.Packet) {
 			b.prevSN = pkt.SequenceNumber - 1
 			b.minTS = pkt.Timestamp - b.maxLate
 			p.reset = true
+		} else {
```
cnderrauber, line 107: does this affect the audio packet?

biglittlebigben:
Not the right fix. The underlying issue was fixed in the SFU.

PR #605, opened 2025-01-24
Add 4 second timeout for each join attempt in region list.

brianbraunstein:


boks1971 (approved):
lgtm!

room.go
```diff
@@ -311,7 +314,15 @@ func (r *Room) JoinWithToken(url, token string, opts ...ConnectOption) error {
...
-				joinRes, err = r.engine.Join(bestURL, token, params)
+				// Not exposing this timeout as an option for now so that callers don't
+				// set unrealistic values.  We may reconsider in the future though.
+				// 4 seconds chosen to balance the trade-offs:
+				// - Too long, users will given up.
+				// - Too short, risk frequently timing out on a request that would have
+				//   succeeded.
+				callCtx, cancelCallCtx := context.WithTimeout(ctx, 4 * time.Second)
+				joinRes, err = r.engine.JoinContext(callCtx, bestURL, token, params)
+				cancelCallCtx()
```
boks1971, line 325: guessing with exponential backoff in room.go, it could still time out. I think it does a bunch of tries. Should we limit that also to be within like 15 seconds total maybe?

boks1971:
> How is this library tested before being released?

I usually test it with `lk` (a. k. a `livekit-cli`), but that is not great for testing this specific path. My changes have been more functional and I could use `lk` to test it out. Maybe, @matkam has notes on how he tested this when he first added this path.

boks1971:
> > > How is this library tested before being released?
> > 
> > 
> > I usually test it with `lk` (a. k. a `livekit-cli`), but that is not great for testing this specific path. My changes have been more functional and I could use `lk` to test it out. Maybe, @matkam has notes on how he tested this when he first added this path.
> 
> That sounds reasonable because I can blackhole clusters by modifying `/etc/hosts` on my local machine.
> 
> Are there instructions or examples anywhere I should follow for doing what you do with `lk` ?

The instructions in the repo is good - https://github.com/livekit/livekit-cli. 

And the help menu of the application also.

boks1971:
> ok, manually tested by blackholing in `/etc/hosts` and found a surprising other place that also needed the context.   after a failed connection, an http request is sent to try to gather info on why it failed.  i just passed it the same ctx and it immediately fails because the deadline is already exceeded, which also more-or-less gives the correct error message (context deadline exceeded)

Oh shoot, forgot about that. My bad. Good catch

engine.go
```diff
@@ -139,8 +140,13 @@ func (e *RTCEngine) SetLogger(l protoLogger.Logger) {
 	}
 }
 
+// Deprecated, use JoinContext.
 func (e *RTCEngine) Join(url string, token string, params *SignalClientConnectParams) (*livekit.JoinResponse, error) {
-	res, err := e.client.Join(url, token, *params)
+	return e.JoinContext(context.TODO(), url, token, params)
+}
+
+func (e *RTCEngine) JoinContext(ctx context.Context, url string, token string, params *SignalClientConnectParams) (*livekit.JoinResponse, error) {
```
davidzhao, line 148: 👍

signalclient.go
```diff
@@ -147,7 +158,7 @@ func (c *SignalClient) connect(urlPrefix string, token string, params SignalClie
 	}
 
 	header := newHeaderWithToken(token)
-	conn, hresp, err := websocket.DefaultDialer.Dial(u.String(), header)
+	conn, hresp, err := websocket.DefaultDialer.DialContext(ctx, u.String(), header)
 	if err != nil {
```
davidzhao, line 162: if the err is due to context cancelled, then we should skip checking `/validate`.

davidzhao: ah just read the above comment stream. feel free to ignore :)

PR #606, opened 2025-01-26
Add a nonce to data messages.

boks1971:
~Using unix nano as nonce.~ 
Use 128-bit UUID bytes as nonce.


roomclient.go
```diff
@@ -132,6 +133,8 @@ func (c *RoomServiceClient) SendData(ctx context.Context, req *livekit.SendDataR
 	if err != nil {
 		return nil, err
 	}
+	// add a nonce to enable receiver to de-dupe
+	req.Nonce = mono.UnixNano()
```
boks1971, line 137: Using monotonic unix nano as nonce, but it probably requires a different mechanism which can be replicated in all server sdks.

paulwe: maybe the high n (48?) bits could be random and the low bits could be a counter?

boks1971: okay, will make a 128 but UUID. Not sure if counter would be useful.

boks1971: Switched to 128-bit UUID in this commit - https://github.com/livekit/server-sdk-go/pull/606/commits/2c9472e8bdb1741e1ce3d0dbe00cae2eb32acc05

paulwe (approved):

PR #615, opened 2025-02-19
apply maxAdjustment to initial track sync

frostbyte73:


pkg/synchronizer/participant.go
```diff
@@ -28,51 +28,54 @@ type participantSynchronizer struct {
...
+	if len(p.senderReports) < len(p.tracks) && time.Since(p.firstReport) < 5*time.Second {
+		return
+	}
+
+	// update ntp start time
+	for ssrc, report := range p.senderReports {
+		if t := p.tracks[ssrc]; t != nil {
+			pts := t.getSenderReportPTS(report)
+			ntpStart := mediatransportutil.NtpTime(report.NTPTime).Time().Add(-pts)
+			if p.ntpStart.IsZero() || ntpStart.Before(p.ntpStart) {
```
boks1971, line 69: may want to log if this adjustment is high, in SFU we drop adjustments that are more than 15 seconds. There is no good metric for how much is too high and should be ignored, but maybe logging large changes would be useful information.

boks1971 (approved):

pkg/synchronizer/participant.go
frostbyte73: We don't make any adjustments more than 15ms. Since data is already going into gstreamer, keeping adjustments smaller than the size of an audio frame keeps things from breaking

frostbyte73: I used to have logs for large adjustments, but they were pretty spammy and not very useful. Now we just log the average and max drift for each track at the end

boks1971: got it, sounds good.

PR #622, opened 2025-03-06
feat(datastreams): initial impl

anunaym14:
Have a bunch of questions added as comments, please check those!

TODO: 
~~- Docs~~ ✅ 
~~- Attachments for text streams~~ ✅ 
~~- Sanitize example~~ ✅ 
~~- Removing deprecated DataPacket.Kind usage~~ ✅ 
~~- https://github.com/livekit/server-sdk-go/pull/622#discussion_r1983855601~~ ✅

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -16,6 +16,7 @@ package lksdk
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
@@ -90,8 +91,9 @@ type RTCEngine struct {
 	OnStreamHeader          func(*livekit.DataStream_Header, string)
 	OnStreamChunk           func(*livekit.DataStream_Chunk)
 	OnStreamTrailer         func(*livekit.DataStream_Trailer)
-	onClose                 []func()
 
+	onClose     []func()
+	onCloseLock sync.Mutex
 	// callbacks to get data
 	CbGetLocalParticipantSID func() string
 }
@@ -192,7 +194,9 @@ func (e *RTCEngine) JoinContext(ctx context.Context, url string, token string, p
 }
 
 func (e *RTCEngine) OnClose(onClose func()) {
+	e.onCloseLock.Lock()
 	e.onClose = append(e.onClose, onClose)
+	e.onCloseLock.Unlock()
 }
 
 func (e *RTCEngine) Close() {
@@ -205,9 +209,11 @@ func (e *RTCEngine) Close() {
 			time.Sleep(50 * time.Millisecond)
 		}
 
+		e.onCloseLock.Lock()
 		for _, onCloseHandler := range e.onClose {
 			onCloseHandler()
 		}
+		e.onCloseLock.Unlock()
 
 		if publisher, ok := e.Publisher(); ok {
 			_ = publisher.Close()
@@ -801,28 +807,30 @@ func (e *RTCEngine) makeRTCConfiguration(iceServers []*livekit.ICEServer, client
 	return configuration
 }
 
-func (e *RTCEngine) publishDataPacket(pck *livekit.DataPacket, kind livekit.DataPacket_Kind) {
+func (e *RTCEngine) publishDataPacket(pck *livekit.DataPacket, kind livekit.DataPacket_Kind) error {
 	data, err := proto.Marshal(pck)
 	if err != nil {
 		e.log.Errorw("could not marshal data packet", err)
-		return
+		return err
 	}
 
 	err = e.ensurePublisherConnected(true)
 	if err != nil {
 		e.log.Errorw("could not ensure publisher connected", err)
-		return
+		return err
 	}
 
 	dc := e.GetDataChannel(kind)
 	if dc == nil {
 		e.log.Errorw("could not get data channel", nil, "kind", kind)
-		return
+		return errors.New("datachannel not found")
 	}
 
 	dc.Send(data)
+	return nil
 }
 
+// TODO: adjust RPC methods to return error on publishDataPacket failure
 func (e *RTCEngine) publishRpcResponse(destinationIdentity, requestId string, payload *string, error *RpcError) {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: []string{destinationIdentity},
@@ -884,7 +892,7 @@ func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, pa
 	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
 }
 
-func (e *RTCEngine) publishStreamHeader(header *livekit.DataStream_Header, destinationIdentities []string) {
+func (e *RTCEngine) publishStreamHeader(header *livekit.DataStream_Header, destinationIdentities []string) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: destinationIdentities,
 		Kind:                  livekit.DataPacket_RELIABLE,
@@ -893,10 +901,10 @@ func (e *RTCEngine) publishStreamHeader(header *livekit.DataStream_Header, desti
 		},
 	}
 
-	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+	return e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
 }
 
-func (e *RTCEngine) publishStreamChunk(chunk *livekit.DataStream_Chunk, destinationIdentities []string) {
+func (e *RTCEngine) publishStreamChunk(chunk *livekit.DataStream_Chunk, destinationIdentities []string) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: destinationIdentities,
 		Kind:                  livekit.DataPacket_RELIABLE,
@@ -905,10 +913,10 @@ func (e *RTCEngine) publishStreamChunk(chunk *livekit.DataStream_Chunk, destinat
 		},
 	}
 
-	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+	return e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
 }
 
-func (e *RTCEngine) publishStreamTrailer(streamId string, destinationIdentities []string) {
+func (e *RTCEngine) publishStreamTrailer(streamId string, destinationIdentities []string) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: destinationIdentities,
 		Kind:                  livekit.DataPacket_RELIABLE,
@@ -919,21 +927,28 @@ func (e *RTCEngine) publishStreamTrailer(streamId string, destinationIdentities 
 		},
 	}
 
-	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+	return e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
 }
 
-func (e *RTCEngine) isBufferStatusLow(kind livekit.DataPacket_Kind) bool {
+func (e *RTCEngine) isBufferStatusLow(kind livekit.DataPacket_Kind) (bool, error) {
 	dc := e.GetDataChannel(kind)
 	if dc != nil {
-		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold()
+		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold(), nil
 	}
-	return false
+	return false, errors.New("datachannel not found")
 }
 
-func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) {
-	// will block forever if dc is nil
-	// probably have a timeout here?
-	for !e.isBufferStatusLow(kind) {
+func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) error {
+	for {
+		low, err := e.isBufferStatusLow(kind)
+		if err != nil {
+			return err
+		}
+		if low {
+			break
+		}
 		time.Sleep(10 * time.Millisecond)
 	}
+
+	return nil
 }
--- a/engine.go
+++ b/engine.go
@@ -87,6 +87,10 @@ type RTCEngine struct {
 	OnRpcRequest            func(callerIdentity, requestId, method, payload string, responseTimeout time.Duration, version uint32)
 	OnRpcAck                func(requestId string)
 	OnRpcResponse           func(requestId string, payload *string, error *RpcError)
+	OnStreamHeader          func(*livekit.DataStream_Header, string)
+	OnStreamChunk           func(*livekit.DataStream_Chunk)
+	OnStreamTrailer         func(*livekit.DataStream_Trailer)
+	onClose                 []func()
 
 	// callbacks to get data
 	CbGetLocalParticipantSID func() string
@@ -128,6 +132,7 @@ func NewRTCEngine() *RTCEngine {
 		e.token.Store(refreshToken)
 	}
 	e.client.OnClose = func() { e.handleDisconnect(false) }
+	e.onClose = []func(){}
 
 	return e
 }
@@ -186,6 +191,10 @@ func (e *RTCEngine) JoinContext(ctx context.Context, url string, token string, p
 	return res, err
 }
 
+func (e *RTCEngine) OnClose(onClose func()) {
+	e.onClose = append(e.onClose, onClose)
+}
+
 func (e *RTCEngine) Close() {
 	if !e.closed.CompareAndSwap(false, true) {
 		return
@@ -194,6 +203,10 @@ func (e *RTCEngine) Close() {
 	go func() {
 		for e.reconnecting.Load() {
 			time.Sleep(50 * time.Millisecond)
+		}
+
+		for _, onCloseHandler := range e.onClose {
+			onCloseHandler()
 		}
 
 		if publisher, ok := e.Publisher(); ok {
@@ -597,6 +610,18 @@ func (e *RTCEngine) handleDataPacket(msg webrtc.DataChannelMessage) {
 				e.OnRpcResponse(msg.RpcResponse.RequestId, nil, fromProto(res.Error))
 			}
 		}
+	case *livekit.DataPacket_StreamHeader:
+		if e.OnStreamHeader != nil {
+			e.OnStreamHeader(msg.StreamHeader, identity)
+		}
+	case *livekit.DataPacket_StreamChunk:
+		if e.OnStreamChunk != nil {
+			e.OnStreamChunk(msg.StreamChunk)
+		}
+	case *livekit.DataPacket_StreamTrailer:
+		if e.OnStreamTrailer != nil {
+			e.OnStreamTrailer(msg.StreamTrailer)
+		}
 	}
 }
 
@@ -857,4 +882,58 @@ func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, pa
 	}
 
 	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) publishStreamHeader(header *livekit.DataStream_Header, destinationIdentities []string) {
+	packet := &livekit.DataPacket{
+		DestinationIdentities: destinationIdentities,
+		Kind:                  livekit.DataPacket_RELIABLE,
+		Value: &livekit.DataPacket_StreamHeader{
+			StreamHeader: header,
+		},
+	}
+
+	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) publishStreamChunk(chunk *livekit.DataStream_Chunk, destinationIdentities []string) {
+	packet := &livekit.DataPacket{
+		DestinationIdentities: destinationIdentities,
+		Kind:                  livekit.DataPacket_RELIABLE,
+		Value: &livekit.DataPacket_StreamChunk{
+			StreamChunk: chunk,
+		},
+	}
+
+	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) publishStreamTrailer(streamId string, destinationIdentities []string) {
+	packet := &livekit.DataPacket{
+		DestinationIdentities: destinationIdentities,
+		Kind:                  livekit.DataPacket_RELIABLE,
+		Value: &livekit.DataPacket_StreamTrailer{
+			StreamTrailer: &livekit.DataStream_Trailer{
+				StreamId: streamId,
+			},
+		},
+	}
+
+	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) isBufferStatusLow(kind livekit.DataPacket_Kind) bool {
+	dc := e.GetDataChannel(kind)
+	if dc != nil {
+		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold()
+	}
+	return false
+}
+
+func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) {
+	// will block forever if dc is nil
+	// probably have a timeout here?
+	for !e.isBufferStatusLow(kind) {
+		time.Sleep(10 * time.Millisecond)
+	}
 }
--- a/engine.go
+++ b/engine.go
@@ -837,6 +837,7 @@ func (e *RTCEngine) publishDataPacketReliable(pck *livekit.DataPacket) error {
 	return e.publishDataPacket(pck, livekit.DataPacket_RELIABLE)
 }
 
+//lint:ignore U1000 Ignore unused function
 func (e *RTCEngine) publishDataPacketLossy(pck *livekit.DataPacket) error {
 	return e.publishDataPacket(pck, livekit.DataPacket_LOSSY)
 }
--- a/engine.go
+++ b/engine.go
@@ -210,10 +210,13 @@ func (e *RTCEngine) Close() {
 		}
 
 		e.onCloseLock.Lock()
-		for _, onCloseHandler := range e.onClose {
+		onClose := e.onClose
+		e.onClose = []func(){}
+		e.onCloseLock.Unlock()
+
+		for _, onCloseHandler := range onClose {
 			onCloseHandler()
 		}
-		e.onCloseLock.Unlock()
 
 		if publisher, ok := e.Publisher(); ok {
 			_ = publisher.Close()
@@ -930,25 +933,16 @@ func (e *RTCEngine) publishStreamTrailer(streamId string, destinationIdentities 
 	return e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
 }
 
-func (e *RTCEngine) isBufferStatusLow(kind livekit.DataPacket_Kind) (bool, error) {
+func (e *RTCEngine) isBufferStatusLow(kind livekit.DataPacket_Kind) bool {
 	dc := e.GetDataChannel(kind)
 	if dc != nil {
-		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold(), nil
+		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold()
 	}
-	return false, errors.New("datachannel not found")
+	return false
 }
 
-func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) error {
-	for {
-		low, err := e.isBufferStatusLow(kind)
-		if err != nil {
-			return err
-		}
-		if low {
-			break
-		}
+func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) {
+	for !e.isBufferStatusLow(kind) {
 		time.Sleep(10 * time.Millisecond)
 	}
-
-	return nil
 }
--- a/engine.go
+++ b/engine.go
@@ -833,11 +833,18 @@ func (e *RTCEngine) publishDataPacket(pck *livekit.DataPacket, kind livekit.Data
 	return nil
 }
 
+func (e *RTCEngine) publishDataPacketReliable(pck *livekit.DataPacket) error {
+	return e.publishDataPacket(pck, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) publishDataPacketLossy(pck *livekit.DataPacket) error {
+	return e.publishDataPacket(pck, livekit.DataPacket_LOSSY)
+}
+
 // TODO: adjust RPC methods to return error on publishDataPacket failure
-func (e *RTCEngine) publishRpcResponse(destinationIdentity, requestId string, payload *string, error *RpcError) {
+func (e *RTCEngine) publishRpcResponse(destinationIdentity, requestId string, payload *string, err *RpcError) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: []string{destinationIdentity},
-		Kind:                  livekit.DataPacket_RELIABLE,
 		Value: &livekit.DataPacket_RpcResponse{
 			RpcResponse: &livekit.RpcResponse{
 				RequestId: requestId,
@@ -845,9 +852,9 @@ func (e *RTCEngine) publishRpcResponse(destinationIdentity, requestId string, pa
 		},
 	}
 
-	if error != nil {
+	if err != nil {
 		packet.Value.(*livekit.DataPacket_RpcResponse).RpcResponse.Value = &livekit.RpcResponse_Error{
-			Error: error.toProto(),
+			Error: err.toProto(),
 		}
 	} else {
 		if payload == nil {
@@ -860,13 +867,16 @@ func (e *RTCEngine) publishRpcResponse(destinationIdentity, requestId string, pa
 		}
 	}
 
-	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+	publishErr := e.publishDataPacketReliable(packet)
+	if publishErr != nil {
+		e.log.Errorw("could not publish rpc response", publishErr)
+	}
+	return publishErr
 }
 
-func (e *RTCEngine) publishRpcAck(destinationIdentity, requestId string) {
+func (e *RTCEngine) publishRpcAck(destinationIdentity, requestId string) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: []string{destinationIdentity},
-		Kind:                  livekit.DataPacket_RELIABLE,
 		Value: &livekit.DataPacket_RpcAck{
 			RpcAck: &livekit.RpcAck{
 				RequestId: requestId,
@@ -874,13 +884,16 @@ func (e *RTCEngine) publishRpcAck(destinationIdentity, requestId string) {
 		},
 	}
 
-	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+	publishErr := e.publishDataPacketReliable(packet)
+	if publishErr != nil {
+		e.log.Errorw("could not publish rpc ack", publishErr)
+	}
+	return publishErr
 }
 
-func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, payload string, responseTimeout time.Duration) {
+func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, payload string, responseTimeout time.Duration) error {
 	packet := &livekit.DataPacket{
 		DestinationIdentities: []string{destinationIdentity},
-		Kind:                  livekit.DataPacket_RELIABLE,
 		Value: &livekit.DataPacket_RpcRequest{
 			RpcRequest: &livekit.RpcRequest{
 				Id:                requestId,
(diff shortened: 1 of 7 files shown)


examples/datastreams/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+	defer func() {
+		senderRoom.Disconnect()
+		receiverRoom.Disconnect()
+	}()
+
+	receiverRoom.RegisterTextStreamHandler("text-receive-iter", func(reader *lksdk.TextStreamReader, participantIdentity string) {
+		logger.Infow("received text stream", "participant", participantIdentity)
+		// Q. If I iterate over the reader directly without a goroutine, I don't receive any dc messages
+		go func() {
```
anunaym14, line 63: This does not make for a good public API design, how can I handle this?

examples/datastreams/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// receiverRoom.RegisterByteStreamHandler("file-receive", func(reader *lksdk.ByteStreamReader, participantIdentity string) {
+	// 	logger.Infow("received file stream", "participant", participantIdentity)
+	// 	go func() {
+	// 		all := reader.ReadAll()
+	// 		os.WriteFile("received.pdf", all, 0644)
+	// 	}()
+	// })
+
+	// Q. If I don't add a delay, I don't receive stream header, I do get chunk and trailer though
```
anunaym14, line 103: Without a ~100ms delay here, I somehow don't get a header, but chunks and trailer arrive as expected.

examples/datastreams/main.go
anunaym14: quoted the wrong thing, but basically, I *need* to iterate over the channel inside a goroutine or I stop receiving messages from the data channel (don't know about other messages, but chunks and trailers don't arrive at all)

streams.go
```diff
@@ -0,0 +1,296 @@
...
+			r.handleChunkReceived(chunk)
+			readOnlyChan <- string(chunk.Content)
+		}
+	}()
+
+	return readOnlyChan
+}
+
+func (r *TextStreamReader) ReadAll() string {
+	// Q: what if part of the data is already read through Read()?
```
lukasIO, line 236: readAll should lock the stream reader. it's either iterate via read or readAll, mixing both is not supported in other impemenations

anunaym14: Got it, thanks @lukasIO!

streams.go
```diff
@@ -0,0 +1,296 @@
...
+type TextStreamWriter struct {
+	*baseStreamWriter[string]
+	Info TextStreamInfo
+}
+
+func newTextStreamWriter(info TextStreamInfo, header *protocol.DataStream_Header, e *RTCEngine, destinationIdentities []string, onProgress *func(progress float64)) *TextStreamWriter {
+	e.publishStreamHeader(header, destinationIdentities)
+
+	onWrite := func(data string) {
+		writeStreamBytes([]byte(data), e, info.Id, destinationIdentities, info.Size, onProgress)
```
lukasIO, line 105: this should have proper utf8 splitting so that we don't split a string at a multi-byte character

lukasIO: there's an edge case where a utf-8 character (e.g. an emoji) is at the exact end of a given stream chunk. 

So this multi-byte character would get split in the middle. I think we still need to avoid that. See https://github.com/livekit/client-sdk-js/blob/main/src/room/utils.ts#L629-L652

anunaym14: Interesting, let me check that up!

examples/datastreams/main.go
lukasIO: did you test this with a local livekit server or against cloud?

anunaym14: cloud (prod, mumbai)

localparticipant.go
```diff
@@ -674,3 +678,140 @@ func (p *LocalParticipant) cleanup() {
...
+	if options.Attributes == nil {
+		options.Attributes = make(map[string]string)
+	}
+
+	info := TextStreamInfo{
+		baseStreamInfo: &baseStreamInfo{
+			Id:       *options.StreamId,
+			MimeType: "text/plain",
+			Topic:    options.Topic,
+			// Q: Is this correct?
```
boks1971, line 697: cc: @lukasIO for confirmation on correctness of this timestamp.

localparticipant.go
```diff
@@ -674,3 +678,140 @@ func (p *LocalParticipant) cleanup() {
...
+	}
+
+	writer := newTextStreamWriter(info, header, p.engine, options.DestinationIdentities, options.OnProgress)
+	p.engine.OnClose(writer.onClose)
+
+	return writer
+}
+
+func (p *LocalParticipant) SendText(text string, options StreamTextOptions) TextStreamInfo {
+	if options.TotalSize == nil {
```
boks1971, line 729: does this need to be a pointer? can it check for 0?

localparticipant.go
```diff
@@ -674,3 +678,140 @@ func (p *LocalParticipant) cleanup() {
...
+	if options.MimeType == "" {
+		mimeType := mime.TypeByExtension(filepath.Ext(filePath))
+		options.MimeType = mimeType
+	}
+
+	writer := p.StreamBytes(options)
+
+	fileBytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, err
```
boks1971, line 808: Is there a need to close the `writer` on this error?

room.go
```diff
@@ -188,6 +192,10 @@ func NewRoom(callback *RoomCallback) *Room {
 		connectionState:    ConnectionStateDisconnected,
 		regionURLProvider:  newRegionURLProvider(),
 		rpcHandlers:        &sync.Map{},
```
boks1971, line 194: :nit: put this at the end to match declaration ☝️ . Makes it easier to read/eye-ball.

room.go
```diff
@@ -1011,6 +1021,120 @@ func (r *Room) handleIncomingRpcRequest(callerIdentity, requestId, method, paylo
 	r.engine.publishRpcResponse(callerIdentity, requestId, &response, nil)
 }
 
+func (r *Room) RegisterTextStreamHandler(topic string, handler TextStreamHandler) error {
+	if _, ok := r.textStreamHandlers.Load(topic); ok {
```
boks1971, line 1025: This is racy in case two goroutines call this. Should not happen, but would be good to avoid these kind of issues. Maybe, just make regular map with a lock and use safe access.

engine.go
```diff
@@ -853,3 +878,57 @@ func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, pa
...
+	dc := e.GetDataChannel(kind)
+	if dc != nil {
+		return dc.BufferedAmount() <= dc.BufferedAmountLowThreshold()
+	}
+	return false
+}
+
+func (e *RTCEngine) waitForBufferStatusLow(kind livekit.DataPacket_Kind) {
+	// will block forever if dc is nil
+	// probably have a timeout here?
```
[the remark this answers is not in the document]
boks1971: Actually return an `error` and bubble that up all the way?

streams.go
```diff
@@ -0,0 +1,296 @@
...
+		close(r.reader)
+		r.closed = true
+	}
+}
+
+type TextStreamReader struct {
+	*baseStreamReader
+	Info TextStreamInfo
+
+	// Q: what are we using this for?
```
boks1971, line 190: cc: @lukasIO for details on how this is used.

examples/datastreams/main.go
boks1971: can any required go routines  be moved to internal implementation? Do not fully understand the flow, but yeah, requiring a goroutine for this is not great.

examples/datastreams/main.go
boks1971: Is this consistently reproducible?

boks1971 (commented):
LGTM! But, I do not understand the nuances of the e2e flow and the protocol. Left some comments.

localparticipant.go
lukasIO: yeah timestamp looks good

streams.go
lukasIO: this was used until a week back to construct the string from chunks that have been received so far, but we got rid of this, so the receivedChunks aren't needed right now. 

We might need them (or something along the lines) once we get to chunk NACKs

examples/datastreams/main.go
anunaym14: yep it is fairly consistent

engine.go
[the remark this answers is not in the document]
anunaym14: > Actually return an error and bubble that up all the way?

Yeah I was thinking the same

localparticipant.go
anunaym14: there's also `time.Now().UnixNano()` so wanted to be sure about which one to use.

localparticipant.go
anunaym14: Good catch

localparticipant.go
anunaym14: yeah fair, will fix this

examples/datastreams/main.go
anunaym14: Can't do that in the current implementation but, we talked about changing the read/write APIs to more closely resemble https://pkg.go.dev/io, will try to handle it there.

localparticipant.go
boks1971: Ideally, we can move to higher resolution timestamps everywhere. If we can use nano everywhere, would be good to do that.

lukasIO: letting clients set the timestamp isn't ideal anyways. JS only has ms, so I'd say we leave it at that for now

examples/datastreams/main.go
anunaym14: This is handled internally now

examples/datastreams/main.go
anunaym14: It was due to DC not being ready on the receiving side

streams.go
anunaym14: In the current implementation with Buffer, the user can use a mix of read methods, I think it's better that way. ReadAll() here returns what is unread.

@boks1971 what do you think?

engine.go
```diff
@@ -181,6 +186,10 @@ func (e *RTCEngine) JoinContext(ctx context.Context, url string, token string, p
 	return res, err
 }
 
+func (e *RTCEngine) OnClose(onClose func()) {
+	e.onClose = append(e.onClose, onClose)
```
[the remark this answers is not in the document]
anunaym14: done

engine.go
```diff
@@ -191,6 +200,10 @@ func (e *RTCEngine) Close() {
 			time.Sleep(50 * time.Millisecond)
 		}
 
+		for _, onCloseHandler := range e.onClose {
```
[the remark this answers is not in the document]
anunaym14: done

engine.go
```diff
@@ -853,3 +878,57 @@ func (e *RTCEngine) publishRpcRequest(destinationIdentity, requestId, method, pa
...
+		},
+	}
+
+	e.publishDataPacket(packet, livekit.DataPacket_RELIABLE)
+}
+
+func (e *RTCEngine) publishStreamChunk(chunk *livekit.DataStream_Chunk, destinationIdentities []string) {
+	packet := &livekit.DataPacket{
+		DestinationIdentities: destinationIdentities,
+		Kind:                  livekit.DataPacket_RELIABLE,
```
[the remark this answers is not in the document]
anunaym14: let me handle this in a subsequent commit once we've finalized other things

engine.go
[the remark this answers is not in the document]
anunaym14: Done, but error propagation is a bit tricky, please check the newly added comments

localparticipant.go
anunaym14: Left it as a pointer as that's what the proto message expects, or should I change it to 0 for the public API and convert internally?

room.go
anunaym14: Used LoadOrStore instead

anunaym14:
@boks1971 please check this again, I'll replied to some of the comments and marked others which are done as resolved. Feel free to mark them as unresolved if you think something needs a change

engine.go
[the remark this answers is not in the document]
anunaym14: Thinking about this again, we should not return an error if the datachannel isn't available, instead we should wait for it, then all the publish calls won't have any reason to fail either, so we can get rid of the error propagation and call this in a goroutine internally.

The above ideology fails if a data channel closes unexpectedly, in that case, do we open a new one?

localparticipant.go
boks1971: not having a pointer makes it nicer to use, but that will one field different from protocol 🤔 . Easier decision if 0 is a valid value. If 0 is not valid, I think keeping it a value and convert to pointer internally would be better.

localparticipant.go
```diff
@@ -674,3 +678,176 @@ func (p *LocalParticipant) cleanup() {
...
+		_, err := writer.Write(text)
+		if err != nil {
+			logger.Errorw("error writing text", err)
+		}
+		err = writer.Close()
+		if err != nil {
+			logger.Errorw("error closing writer", err)
+		}
+	}()
+	return &writer.Info, nil
```
boks1971, line 758: How is the caller expected to use this return value? What if there is an error in writing (guessing closing error is not problematic)? Should the application retry? Is that what the info return value will be used for? In the example, I see that the return value is not used at all.

Have to use callback to notify errors when doing async send. Or use await semantics. There is a comment there about needing to return error, is that a TODO?

localparticipant.go
```diff
@@ -674,3 +678,176 @@ func (p *LocalParticipant) cleanup() {
...
+			textHeader.TextHeader.ReplyToStreamId = *options.ReplyToStreamId
+		}
+	}
+
+	writer, err := newTextStreamWriter(info, header, p.engine, options.DestinationIdentities, options.OnProgress)
+	if err != nil {
+		return nil, err
+	}
+	p.engine.OnClose(func() {
+		err := writer.Close()
```
boks1971, line 726: A couple of bits here (assuming I am reading the code right)
1. The write goroutine below seems to keep going even if this closed.
2. On close, the streams is trying to write end chunk. Two things there
     a. Does it make sense to write end chunk if only partial was written?
     b. Looks like close could write an end chunk and then the write can continue to write more?

boks1971: Same concerns also for `SendFile` paths below.

room.go
```diff
@@ -997,7 +1006,7 @@ func (r *Room) handleIncomingRpcRequest(callerIdentity, requestId, method, paylo
 		if _, ok := err.(*RpcError); ok {
 			r.engine.publishRpcResponse(callerIdentity, requestId, nil, err.(*RpcError))
 		} else {
-			r.engine.log.Warnw("Uncaught error returned by RPC handler for method, using application error instead", err, "method", method)
+			r.log.Warnw("Uncaught error returned by RPC handler for method, using application error instead", err, "method", method)
```
boks1971, line 1009: :nit: lower case `u` in `Uncaught`, our convention is to use lower case in logging.

room.go
```diff
@@ -1011,6 +1020,118 @@ func (r *Room) handleIncomingRpcRequest(callerIdentity, requestId, method, paylo
...
+func (r *Room) handleStreamTrailer(streamTrailer *livekit.DataStream_Trailer) {
+	streamId := streamTrailer.StreamId
+
+	byteStreamReader, ok := r.byteStreamReaders.Load(streamId)
+	if ok {
+		reader := byteStreamReader.(*ByteStreamReader)
+		for k, v := range streamTrailer.Attributes {
+			reader.Info.Attributes[k] = v
+		}
+		reader.close()
```
boks1971, line 1120: The reader is closed immediately after setting the attributes. What is the purpose of it? Who uses those attributes?

streams.go
```diff
@@ -0,0 +1,311 @@
...
+		end := i + STREAM_CHUNK_SIZE
+		if end > len(data) {
+			end = len(data)
+		}
+
+		// this is a blocking call, but, if we call it in a goroutine internally,
+		// error propagation will require some blocking code which ruins the point
+		// is it best for the user to call this in a goroutine?
+		// or do we not block on buffer status low and just let the user handle it?
+		err := e.waitForBufferStatusLow(protocol.DataPacket_RELIABLE)
```
boks1971, line 88: no, this should not be exposed to the user. I think (if I am reading the code right), this whole function is getting called from a goroutine, so blocking here should be fine, but need to ensure that any error flows back to the user.

streams.go
```diff
@@ -0,0 +1,311 @@
...
+			StreamId:   streamId,
+			Content:    data[i:end],
+			ChunkIndex: chunkIndex,
+		}, destinationIdentities); err != nil {
+			return bytesWritten, err
+		}
+
+		if onProgress != nil && totalSize != nil {
+			progress := float64(i) / float64(*totalSize)
+			(*onProgress)(progress)
```
boks1971, line 103: Wondering if `onProgress` can be used to indicate error too. Is `onProgress` going back to the user?

streams.go
```diff
@@ -0,0 +1,311 @@
...
+	r.maybeCallOnProgress(len(n))
+	return n, err
+}
+
+func (r *TextStreamReader) ReadAll() string {
+	// wait for the stream to be closed
+	// not sure if this is the best way to do this
+	for {
+		r.closedLock.Lock()
+		if r.closed {
```
boks1971, line 265: can `closed` be atomic to avoid all the extra locking just for that one variable. Maybe, `readBuffer.Write` ☝️ requires a lock?

streams.go
```diff
@@ -0,0 +1,311 @@
...
+func NewByteStreamReader(info ByteStreamInfo, totalChunkCount *uint64) *ByteStreamReader {
+	return &ByteStreamReader{
+		baseStreamReader: newBaseStreamReader(totalChunkCount),
+		Info:             info,
+	}
+}
+
+func (r *ByteStreamReader) ReadAll() []byte {
+	// wait for the stream to be closed
+	// not sure if this is the best way to do this
```
boks1971, line 292: cannot think of a better way, guess you have to wait for close to read all. Maybe, can use something like https://github.com/livekit/protocol/blob/549a29a812048270a3726acd7940974231d8b947/utils/promise.go#L15 to make it easier to do this where the promise resolves on `close`.

boks1971 (commented):
Thank you for addressing comments @anunaym14 . I think there are still bits that need some clarification. Have left some comments.

localparticipant.go
anunaym14: Got it

engine.go
```diff
@@ -191,6 +204,12 @@ func (e *RTCEngine) Close() {
 			time.Sleep(50 * time.Millisecond)
 		}
 
+		e.onCloseLock.Lock()
+		for _, onCloseHandler := range e.onClose {
+			onCloseHandler()
+		}
+		e.onCloseLock.Unlock()
```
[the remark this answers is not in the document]
anunaym14: Right, my bad

examples/datastreams/main.go
```diff
@@ -0,0 +1,137 @@
...
+		receiverRoom.Disconnect()
+	}()
+
+	receiverRoom.RegisterTextStreamHandler("text-receive-iter", func(reader *lksdk.TextStreamReader, participantIdentity string) {
+		logger.Infow("received text stream", "participant", participantIdentity)
+		all := reader.ReadAll()
+		logger.Infow("received text", "text", all)
+	})
+
+	// receiverRoom.RegisterTextStreamHandler("text-receive-all", func(reader *lksdk.TextStreamReader, participantIdentity string) {
```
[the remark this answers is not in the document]
anunaym14: Yeah, tests are useful. I'm waiting for a final public API design before sanitizing them

streams.go
anunaym14: OnProgress is optional, user might not set it

streams.go
anunaym14: it's being called from a goroutine for methods like SendText and SendFile, for StreamText and StreamBytes, where the user calls Write() themselves, this is an issue.

I'm not talking about exposing it in the comment, I was talking about not waiting for the buffer status to be low in our code, and returning an error instead, the user can then handle it themselves. But I'm not sold on this fully, maybe an error callback would be a better idea.

streams.go
anunaym14: Right yea that should be behind a lock

room.go
anunaym14: I'm not sure about this, I saw the JS SDK doing it. Maybe @lukasIO can help.

localparticipant.go
anunaym14: yeah that is a todo, wasn't sure how to propagate the error without blocking anything. maybe callback is the way.

room.go
boks1971: I see that `ReadAll()` waits for close. So, I guess this is a signal for `ReadAll` to unblock and read?

localparticipant.go
anunaym14: > Does it make sense to write end chunk if only partial was written?

I think so, end chunk is essential for the other user to close the reader 

> Looks like close could write an end chunk and then the write can continue to write more?

Good catch, let me handle this.

streams.go
boks1971: Not sure about mixed methods. Need to think about it more.

room.go
anunaym14: yeah that's for close yeah, I'm not sure what the trailer attributes are for

lukasIO: JS uses the inherent mutability of objects to update attributes on the stream's info with additional attributes (or overrides) that are sent in the trailer. 
The idea was that e.g. for transcriptions when you're beginning a stream only the `startTime` is known. Only once the stream is finishing you'll have knowledge about the `endTime` so we allowed for updating attributes also as part of the trailer. 

In JS accessing `attributes` before and after a stream has closed can have different attributes set for that reason

engine.go
```diff
@@ -205,10 +205,13 @@ func (e *RTCEngine) Close() {
 		}
 
 		e.onCloseLock.Lock()
-		for _, onCloseHandler := range e.onClose {
+		onClose := e.onClose
+		e.onClose = []func(){}
```
boks1971, line 209: Is there a need to clear `onClose`. Guess it can be left as is right?

examples/datastreams/main.go
```diff
@@ -81,13 +82,13 @@ func main() {
...
-	// })
+	receiverRoom.RegisterTextStreamHandler("text-stream-recv-all", func(reader *lksdk.TextStreamReader, participantIdentity string) {
+		logger.Infow("received text stream", "participant", participantIdentity)
+		go func() {
+			all := reader.ReadAll()
+			logger.Infow("received full text chunks", "text", all)
+		}()
+	})
 
 	// receiverRoom.RegisterByteStreamHandler("file-receive", func(reader *lksdk.ByteStreamReader, participantIdentity string) {
```
boks1971, line 93: If you want to leave these commented out, maybe adding a `TODO` would be useful to make sure this either gets implemented or cleaned up later. Here are a couple of simple rules I use
1. Six months from now, if I look at this and if this is commented out, will I remember why it is commented out? If I don't put notes, I will not. So, I usually delete it (make a task elsewhere to implement these) or put a TODO here explaining what is this test case for and why it is not implemented. So, six months down the line, I will have some clue.
2. I typically use a more feature specific TODO. For this, I would do something like `DATASTREAM-TODO`. So, I can just search for that later. If I do simple TODO, they just build up over time and become an issue to sift through. So, I tend to use TODO markers with feature name attached to it.

boks1971 (approved):

examples/datastreams/main.go
anunaym14: Yeah, sanitizing the example is a TODO!

engine.go
anunaym14: Yeah it can be left as is as well, but also no harm in cleaning these up instead of waiting for the go gc, since these won't be used again

streams.go
anunaym14: This is done in 1723103e5e35964881d8c514d4d9fa9952c4f536, please have a look.

cc @boks1971

examples/datastreams/main.go
```diff
@@ -59,67 +81,54 @@ func main() {
...
+		res := ""
+		for {
+			word, err := reader.ReadString(' ')
+			logger.Infow("read word", "word", word)
+			res += word
+			if err != nil {
+				// EOF represents the end of the stream
+				if err == io.EOF {
+					break
+				} else if err == lksdk.EAGAIN {
```
boks1971: Are you leaving this to the user? I thought we could just do it internally and externally, the user either gets data or io.EOF. WDYT?

examples/datastreams/main.go
```diff
@@ -139,11 +148,13 @@ func main() {
...
-	// })
+	downloadTestFile()
+
+	topic = "file-read-all"
+	senderRoom.LocalParticipant.SendFile(testFileName, lksdk.StreamBytesOptions{
+		Topic:    topic,
+		FileName: &testFileName,
+	})
 
 	time.Sleep(30 * time.Second)
```
boks1971, line 163: Maybe use a wait group to ensure that all the stuff and exit rather than sleep?

streams.go
```diff
@@ -127,28 +127,25 @@ func (w *baseStreamWriter[T]) Close() {
...
+			Content:    chunk,
 			ChunkIndex: chunkIndex,
 		}, w.destinationIdentities); err != nil {
 			w.engine.log.Errorw("could not publish stream chunk", err)
 		}
 
 		if w.onProgress != nil && w.totalSize != nil {
-			progress := float64(i) / float64(*w.totalSize)
+			progress := float64(len(chunk)) / float64(*w.totalSize)
 			(*w.onProgress)(progress)
```
boks1971, line 192: Don't think you need to de-ref here, you can just call `w.onProgress(progress)` I believe.

streams.go
```diff
@@ -366,3 +363,46 @@ func (r *ByteStreamReader) ReadAll() []byte {
 type TextStreamHandler func(reader *TextStreamReader, participantIdentity string)
 
 type ByteStreamHandler func(reader *ByteStreamReader, participantIdentity string)
+
+// ---------------------------------------------------------
+
+func chunkUtf8String(s string) [][]byte {
```
boks1971, line 435: would be good to have some UT and also some comments as to what this is doing would be great.

streams.go
```diff
@@ -366,3 +363,46 @@ func (r *ByteStreamReader) ReadAll() []byte {
...
+	chunks := [][]byte{}
+	stringBytes := []byte(s)
+
+	for len(stringBytes) > STREAM_CHUNK_SIZE {
+		k := STREAM_CHUNK_SIZE
```
boks1971, line 440: should this be `k := STREAM_CHUNK_SIZE - 1` , i. e. indexing into last element of `STREAM_CHUNK_SIZE` array?

examples/datastreams/main.go
anunaym14: Right yea, I missed that while cleaning up. Let me do that.

streams.go
anunaym14: It wasn't allowing me to do that

examples/datastreams/main.go
boks1971: Understood, but feels like it can be handled internally.

> notify that to the user with ErrAgain (better to return what we have than waiting for more)
I am not understanding. I think this can return (num_bytes_read, nil) and user would take those bytes and read again.

Internally, if we get (o, io.EOF) from Go APIs before the stream end, we just loop and go back and block on read?

streams.go
```diff
@@ -20,21 +20,46 @@ type baseStreamInfo struct {
 	Attributes map[string]string
 }
 
+// Info for byte streams
+// - Id is the id of the stream
+// - MimeType is the mime type of the stream, calculate for SendFile if not provided
```
boks1971, line 25: :nit: `calculate` -> `determined` (guessing there is no `calculation` per se)

examples/datastreams/main.go
anunaym14: So just on (0, io.EOF)?

In cases of (n, io.EOF) where n > 0, do we return ErrAgain with n? Will modify the error message to day "not enough bytes available"

localparticipant.go
```diff
@@ -726,13 +728,63 @@ func (p *LocalParticipant) SendText(text string, options StreamTextOptions) *Tex
...
 
+	for i, attachment := range options.Attachments {
+		onProgress := func(progress float64) {
+			handleProgress(progress, i+1)
+		}
+		p.SendFile(attachment, StreamBytesOptions{
+			Topic:                 options.Topic,
+			DestinationIdentities: options.DestinationIdentities,
+			StreamId:              &attachedStreamIds[i],
+			OnProgress:            &onProgress,
```
boks1971, line 783: oh you are taking address of this, I think it can just be of type `func` which will be a pointer.

streams.go
anunaym14: If I understand correctly, the idea is that we do not want to split at a continuation byte. So if `k` starts from  `STREAM_CHUNK_SIZE`, and we determine if it is safe to split there (i.e. the string before it is a valid utf-8 string).

When we break out of the loop at `k == STREAM_CHUNK_SIZE` i.e. in the first iteration itself, we get to use the full chunk size available to us. The chunk is formed as `stringBytes[:k]` meaning the `k`th element is not included. 

If we start `k` from `STEAM_CHUNK_SIZE - 1`, that would theoretically work as well, but it's not maxing out on the available max chunk size.

examples/datastreams/main.go
boks1971: Maybe my understanding is incomplete, but this what I am thinking

user
```
for {
    if chunk, err := stream.Read(); err != nil {
      // done
      break
    } else {
       data += chunk
    }
```

internally
```
func Read() ([]byte, error) {
     buf, err := buffer.Read()
     // unclear to me how the end of stream is known, assuming that is known
     if end-of-stream {
          return buf, io.EOF
     }. else if len(buf) > 0 {
         return buf, nil
     } else {
          // len(buf) == 0 and err is probably io.EOF
          time.Sleep(10 * time.Millisecond)
     }
}
```
Effectively, `Read()` blocks till there is data or end of stream.

streams.go
boks1971: got it, makes sense

localparticipant.go
anunaym14: Hmm yes you're right, I forgot about the zero-init here, func type will zero-init with `nil` if not passed. Will update this, thanks.

examples/datastreams/main.go
anunaym14: Understood. Just one doubt:

The Read() method takes []byte say `p` and returns `len(p)` bytes from the buffer, if buffer has less bytes, it returns whatever is available with `err == nil` and only returns io.EOF when the buffer has nothing to read from.

But, the ReadBytes() method takes a delim byte, and returns till the delim including the delim itself. But, if delim isn't present in the buffer, it returns everything available but with io.EOF.

How should we handle this? Do we return nil in these cases (before stream closure), or do we return either or ErrAgain or io.EOF?

anunaym14: Please check the latest commit, this should be handled properly now!

streams_chunks_test.go
```diff
@@ -0,0 +1,412 @@
...
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+	"unicode/utf8"
+)
+
+func TestChunkUtf8String(t *testing.T) {
+	t.Run("Empty string", func(t *testing.T) {
```
boks1971, line 11: :nit: lower case `e` in `Empty`. Conventionally, we use all lower case for test names, comments, etc. Other tests in the rest of the file too.

streams_chunks_test.go
```diff
@@ -0,0 +1,412 @@
...
+		chunks := chunkUtf8String("")
+		if len(chunks) != 0 {
+			t.Errorf("Expected 0 chunks for empty string, got %d", len(chunks))
```
boks1971, line 14: Does this fail the test suite? We use `require` package for testing. So, this would be something like `require.Equal(t, 0, len(chunks))` (maybe `require` has a convenience method called `require.Zero` as they have a lot of convenience methods). Would be good to stay consistent in testing.

streams_chunks_test.go
```diff
@@ -0,0 +1,412 @@
...
+				t.Errorf("Chunk %d expected size %d, got %d", i, STREAM_CHUNK_SIZE, len(chunks[i]))
+			}
+		}
+
+		// Verify the last partial chunk size
+		if len(chunks[fullChunks]) != extraBytes {
+			t.Errorf("Last chunk expected size %d, got %d", extraBytes, len(chunks[fullChunks]))
+		}
+	})
+}
```
boks1971, line 412: Nice tests!!!

streams.go
```diff
@@ -238,6 +233,12 @@ type baseStreamReader struct {
 
 	closed atomic.Bool
 	lock   sync.Mutex
+	cond   *sync.Cond
+
+	// readerMutex is used to lock the reader when reading the stream
+	// this will prevent issues when the user tries to read from multiple methods at the same time
+	// since cond unlocks the lock before waiting, this is needed to prevent multiple read calls from fighting for data
```
boks1971, line 240: can remove this comment, this feels like comment on how a condition variable works and not specific to this use case, would be good to remove non-specific comments.

streams.go
```diff
@@ -279,49 +282,74 @@ func (r *baseStreamReader) maybeCallOnProgress(n int) {
 	}
 }
 
+// waits for a write to the stream
+func (r *baseStreamReader) waitForData() {
+	// if stream is closed, the methods will return io.EOF automatically
+	for r.readBuffer.Len() == 0 && !r.closed.Load() {
+		// cond.Wait() unlocks the lock before waiting, allowing enqueue to write to the buffer
```
boks1971, line 289: here as well, do not need a comment on how condition variables work.

streams.go
```diff
@@ -279,49 +282,74 @@ func (r *baseStreamReader) maybeCallOnProgress(n int) {
...
-// If the buffer has no data to return, err is ErrAgain or io.EOF (unless len(p) is zero) depending on if the stream is closed or not.
-// Otherwise, the error is nil.
+// If the buffer has no data to return, it will wait for a write to the stream or return io.EOF if the stream is closed.
 func (r *baseStreamReader) Read(bytes []byte) (int, error) {
+	r.readerMutex.Lock()
 	r.lock.Lock()
-	defer r.lock.Unlock()
+	defer func() {
+		r.lock.Unlock()
+		r.readerMutex.Unlock()
```
boks1971, line 314: I think this should be a single mutex for reading/writing and the condition variable attached to that. Don't think there is need for two. Increases chances of deadlocks. Is there a need for two? Condition variable is designed to synchronise accesses like these.

streams.go
```diff
@@ -373,18 +417,20 @@ func (r *TextStreamReader) ReadString(delim byte) (string, error) {
...
+	}()
+
 	// wait for the stream to be closed
-	for {
-		if r.closed.Load() {
-			break
-		}
-		time.Sleep(10 * time.Millisecond)
+	for !r.closed.Load() {
+		// cond.Wait() unlocks the lock before waiting, allowing enqueue to write to the buffer
```
boks1971, line 429: generic comment remove, maybe in more places too.

streams_chunks_test.go
anunaym14: Yeah it fails the test, but let me update these to be consistent with other tests

streams.go
anunaym14: So it was added because if user calls Read() twice and the buffer is empty:

- Read1 acquires the lock as it is called first
- It does to waitForData() and cond.Wait() releases the lock
- Read2 acquires the lock and goes into waitForData() as well
- Now, when the condition is met, we have no way to guarantee which read call returns the data 

Using readerLock will ensure only one read is processed at a time. Without the sync.Cond, just `lock` would've been enough

boks1971: I don't think `Read2` can get the lock. It will block on the top level `lock` as it is written today, right? And the top level `lock` is held while calling `waitForData()` which is not good as it could be locked for a long time.

If there are multiple readers, that would mean the user is doing some reconciliation to put chunks in order and it is their responsibility. From here, does not matter which one returns. It only needs to be consistent, i. e. when a call returns, it either returns with some data (chunk) so that user can piece it together or `io.EOF` when it is ended. Yes it is possible that the two race and `io.EOF` could be returned before the routine returns some valid data, but those are not cases this needs to protect against.

The API contract is
- if you call `Read()`, it will block till there is data or till is stream is closed
- if data returned, use it (up to user on how they want to use it)
- if io.EOF returned, stream is ended

anunaym14: > I don't think Read2 can get the lock. It will block on the top level lock as it is written today, right?

cond.Wait() will unlock the lock acquired by read1 while it waits, so read2 will get a lock

> From here, does not matter which one returns. It only needs to be consistent, i. e. when a call returns, it either returns with some data (chunk) so that user can piece it together or io.EOF when it is ended.

Yeah, it would be consistent before the readerLock too, which ever reader gets the lock first will return the data, if there's more, other reader will return it, or the other reader will start waiting again.

anunaym14: But if there are a bunch of read calls, called inside unique goroutines, we can't guarantee that they'll execute sequentially.

Cool yeah, makes sense to get rid of it.

boks1971: > cond.Wait() will unlock the lock acquired by read1 while it waits, so read2 will get a lock

I am referring to `r.lock.Lock()` which can be considered the `write lock` on this buffer. So, `Read()` acquires both locks and `cond.Wait()` will unlock only the `readerMutex`, but the `write lock` is still held, right? Am I reading it wrong?

> But if there are a bunch of read calls, called inside unique goroutines, we can't guarantee that they'll execute sequentially.

I don't think this should be part of the API contract. By definition, multiple goroutines means you want concurrency and it should be up to the user to manage that concurrency.

anunaym14: > I am referring to r.lock.Lock() which can be considered the write lock on this buffer. So, Read() acquires both locks and cond.Wait() will unlock only the readerMutex, but the write lock is still held, right? Am I reading it wrong?

cond is initiated with the `r.lock` which is the write lock, which is what allows it to write while waiting.`r.readerLock` is still locked.

anunaym14: Pushed a commit  ([99980bd](https://github.com/livekit/server-sdk-go/pull/622/commits/99980bda537ddb1019351fe82f414d2aa00821bb)) to remove this

streams_chunks_test.go
anunaym14: Done! [ea551c5](https://github.com/livekit/server-sdk-go/pull/622/commits/ea551c5ae0aba0f35a1fbc5d920c74f376f8334b)

streams_chunks_test.go
```diff
@@ -83,24 +67,16 @@ func TestChunkUtf8String(t *testing.T) {
 
 		chunks := chunkUtf8String(testString)
 
-		if len(chunks) != 2 {
-			t.Errorf("Expected 2 chunks, got %d", len(chunks))
-		}
+		require.Len(t, chunks, 2)
```
boks1971, line 70: quick note on use on `require`. The convention is to have the `expected` value as the second argument and the `actual` value as the third argument. Can you please have a look and ensure that convention is followed. Sorry for nit-picking, but just want to keep it consistent (I am probably guilty of not following this in several places, but I am correcting things as and when I see it :-). )

anunaym14: You're right, that is the convention for `require.Equal()`, which is used correctly, I've double-checked.

- `require.Zero()` and `require.True()` only take 1 value that they match with zero/true, so that is correct as well.
- That leaves us with `require.Len()` takes an object type that supports `len()` as second parameter and the expected length as the third: https://pkg.go.dev/github.com/stretchr/testify/require#Len

boks1971: Oh cool, did not know that Len has a different convention. Thank you 🙏

examples/datastreams/main.go
davidzhao, line 1: nice!

davidzhao (commented):
lgtm! good work

PR #630, opened 2025-04-08
fix: clear sync.Map properly

anunaym14:


localparticipant.go
```diff
@@ -545,8 +545,10 @@ func (p *LocalParticipant) handleParticipantDisconnected(identity string) {
 
 	p.rpcPendingResponses.Range(func(key, value interface{}) bool {
```
boks1971, line 546: Not necessarily for this PR.

But, something to think about in general.

As this is keyed off of `identity`, what happens is the participant leaves and re-joins very quickly? Could the delete here race with some other place adding it back? This is not a sync.Map issue. Could happen with regular maps also. It is an ordering issue.

room.go manages `remoteParticipants` using both SID and identity. Don't remember the full details, but wondering if some SID based approach should be adopted for this also to deal with quick disconnects/reconnects.

We actually had an issue of a user disconnecting and reconnecting in the same second. That triggered an error because the join time had 1 second resolution (we have since added ms resolution) and that got squashed and it missed the disconnect -> reconnect transition. That problem was in the SFU, but was seen by a Go SDK client as missing participant update.

So, just a general note about using identity and having to deal with very quick disconnect/reconnect.

boks1971 (approved):
lgtm! one note about use of `identity` with reconnects for later consideration.

PR #635, opened 2025-04-18
fix(localtrack): add missing case for h265 payloader

andresmanelli:
Hello, I am testing H265 support and I found this missing case to dispatch the payloader.


cnderrauber (approved):
looks good!

andresmanelli:
@cnderrauber @davidzhao Thanks! Quick question though: is it possible in the sdk to add H265 as a codec for the SDP negotiation?

I had to add it to the default codecs list (c.f. https://github.com/pion/webrtc/pull/3102) so that the negotiation works.

PR #656, opened 2025-05-12
feat(audiotrack): lockless buffer

anunaym14:
Noticing some crackling noise, debugging that. The audio otherwise seems to be in the correct order.

### DIFF
--- a/pkg/lockless_circular_buffer/circular_buffer.go
+++ b/pkg/lockless_circular_buffer/circular_buffer.go
@@ -1,0 +1,330 @@
+package lockless_circular_buffer
+
+import (
+	"runtime"
+	"sync/atomic"
+	"time"
+	"unsafe"
+)
+
+type CircularBuffer[T any] struct {
+	buffer []T
+	head   uint32
+	tail   uint32
+	mask   uint32
+	size   uint32
+
+	// todo(anunaym14): arch-aware padding to avoid false sharing
+	// _padding [x]byte
+}
+
+func NewCircularBuffer[T any](capacity uint32) *CircularBuffer[T] {
+	// Ensure capacity is a power of 2 and at least 1
+	// todo(anunaym14): cleanup
+	if capacity == 0 {
+		capacity = 1
+	} else if (capacity & (capacity - 1)) != 0 {
+		capacity--
+		capacity |= capacity >> 1
+		capacity |= capacity >> 2
+		capacity |= capacity >> 4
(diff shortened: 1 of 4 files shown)


pkg/lockless_circular_buffer/circular_buffer.go
```diff
@@ -0,0 +1,330 @@
+package lockless_circular_buffer
+
+import (
+	"runtime"
+	"sync/atomic"
```
boks1971, line 5: We standardise on uber atomic. It is nicer to use.

anunaym14: Gotcha, will replace.

pkg/lockless_circular_buffer/circular_buffer.go
```diff
@@ -0,0 +1,331 @@
...
+	if capacity == 0 {
+		capacity = 1
+	} else if (capacity & (capacity - 1)) != 0 {
+		capacity--
+		capacity |= capacity >> 1
+		capacity |= capacity >> 2
+		capacity |= capacity >> 4
+		capacity |= capacity >> 8
+		capacity |= capacity >> 16
+		capacity++
```
paulwe, line 34: bits.Len32

pkg/lockless_circular_buffer/circular_buffer.go
```diff
@@ -0,0 +1,331 @@
...
+	"time"
+	"unsafe"
+
+	"go.uber.org/atomic"
+)
+
+type CircularBuffer[T any] struct {
+	buffer []T
+	head   *atomic.Uint32
+	tail   *atomic.Uint32
```
paulwe, line 14: these are zero initialized automatically so you can avoid the extra indirection and skip initializing with New(0)

pkg/lockless_circular_buffer/circular_buffer.go
```diff
@@ -0,0 +1,331 @@
...
+			currentBatchSize = availableItems
+		}
+
+		newHead := (head + currentBatchSize) & cb.mask
+
+		if cb.head.CompareAndSwap(head, newHead) {
+			for i := uint32(0); i < currentBatchSize; i++ {
+				slotIndex := (head + i) & cb.mask
+				result = append(result, cb.buffer[slotIndex])
+			}
```
paulwe, line 268: this can be implemented in at most 2 copies? after head has been updated is the read range safe from concurrent writes?

PR #671, opened 2025-05-25
Keep chunks in deque.

boks1971:
Re-jiggering a bit of this. Using more standard definitions

sample: a single qunatised digital audio sample (a single point in time)
frame: a collection of samples with some well-defined duration - 10 ms used here for Opus encode (side note: I wish the writer API was called WriteFrame instead of WriteSample)
chunk: a collection of samples of variable duration - like what is submitted on PCM side, could be 1 second long, could be 100 ms long, etc.
Change the deque to store chunks as is and when processing construct 10 ms frames. Don't think there is a need to convert to samples while storing in deque. Should keep the deque size smaller.

### DIFF
--- a/examples/audiotrack/main.go
+++ b/examples/audiotrack/main.go
@@ -138,7 +138,7 @@ func handleSubscribe(track *webrtc.TrackRemote, targetChannels int) (*lkmedia.PC
 		panic(err)
 	}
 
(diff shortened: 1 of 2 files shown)


anunaym14 (approved):
LGTM, thanks!

PR #674, opened 2025-05-27
feat(PCMTrack): e2ee support

anunaym14:


examples/audiotrack/main.go
```diff
@@ -132,14 +138,40 @@ func handlePublish(room *lksdk.Room) {
...
 
-func handleSubscribe(track *webrtc.TrackRemote, targetChannels int) (*lkmedia.PCMRemoteTrack, *os.File) {
+func handleSubscribe(track *webrtc.TrackRemote, publication *lksdk.RemoteTrackPublication, targetChannels int) (*lkmedia.PCMRemoteTrack, *os.File) {
 	fileWriter, err := os.Create("test.mka")
 	if err != nil {
 		panic(err)
 	}
 
+	var decryptor lkmedia.Decryptor
+	passphrase := "helloworld"
```
lukasIO, line 148: this seems to imply a static key for the lifetime of a subscription. 

I'm not sure if it's necessary here to implement key updates for go, just wanted to call out the constraints that this imposes.

I think that also in turn means there's currently no support for the key index defined in the encryption trailer in this SDK?

anunaym14: Please check: [03b9830](https://github.com/livekit/server-sdk-go/pull/674/commits/03b983023048e7674b723bbea919706cc2761d94)

anunaym14: > I think that also in turn means there's currently no support for the key index defined in the encryption trailer in this SDK?

Let me look into this

boks1971 (approved):

examples/audiotrack/main.go
anunaym14: Had a look, `kid` is not used for decrypting, the method directly expects the user to use the right key from the ring (bad design).  It is passed as parameter for encrypting though, so key index is technically supported, but totally managed by the user.

pkg/media/audiotrack.go
```diff
@@ -379,6 +404,17 @@ func NewPCMRemoteTrack(track *webrtc.TrackRemote, writer PCMRemoteTrackWriter, o
...
+	// Should it be done before or after jitter handling?
+	// seems to work either way.
+	// Currently it's after jitter handling.
+	if t.decryptor != nil {
+		// Before decrypting, I would like to check here if
+		// the decryption handler is handling the correct type of
+		// encryption, but, couldn't figure out a way without making the public API messy.
+		h = newDecryptionHandler(h, t.decryptor)
+	}
+
```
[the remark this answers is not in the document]
anunaym14: Improved the comment, the option was to pass RemoteTrackPublication as part of the options but it does not force the user to pass it.

We could also pass it while building the Encryptor/Decryptor but since those are generic and not related to the track, it did not feel right to pass it there. So left it to be user-managed for now.

PR #677, opened 2025-05-29
Remove option to not write silence on no data.

boks1971:
Although a nice option, unfortunately it does not work with media-sdk writer pipeline. The writer pipeline submits a sample into the track's packetiser with just the duration of the sample. So, if samples are not provided periodically, the RTP time stamp calculation and therefore the RTCP Sender Report time stamp calculation will both be wrong.

We can expose this option when the pipeline can accommodate gaps. Till then, it is not good to have this in the public API as it would lead to incorrect use/behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,12 +250,6 @@ if err != nil {
 
 The SDK will encode the sample to Opus and write it to the track. If the sourceSampleRate is not 48000, resampling is also handled internally.
 
-The API also provides an option to write silence to the track when no data is available, this is disabled by default but you can enable it using:
-
-```go
-publishTrack, err := lkmedia.NewPCMLocalTrack(sourceSampleRate, sourceChannels, logger.GetLogger(), lkmedia.WithWriteSilenceOnNoData(true))
-```
-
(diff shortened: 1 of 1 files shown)


pkg/media/audiotrack.go
```diff
@@ -25,18 +25,11 @@ const (
...
 type PCMLocalTrackParams struct {
-	WriteSilenceOnNoData bool
-	Encryptor            Encryptor
+	Encryptor Encryptor
 }
 
 type PCMLocalTrackOption func(*PCMLocalTrackParams)
 
-func WithWriteSilenceOnNoData(writeSilenceOnNoData bool) PCMLocalTrackOption {
-	return func(p *PCMLocalTrackParams) {
```
anunaym14, line 35: Should we mark this deprecated instead of removing it? 

But, we're altering the behavior, so removing is probably better.

anunaym14 (approved):
LGTM! 

The ReadMe will also need changes to reflect this

pkg/media/audiotrack.go
boks1971: Yeah, removing is better. There is only one user now. We can request them to remove that call. 

Will change README.

paulwe (approved):

anunaym14:
https://github.com/livekit/server-sdk-go/pull/678

here it is

PR #680, opened 2025-05-30
feat(PCMLocalTrack): support muting

anunaym14:
Please review the changes commit by commit, here's a summary:

- Fix typo on the encryptor file name
- Split up PCMLocalTrack and PCMRemoteTrack to different files
- Add muting support to PCMLocalTrack


private.go
```diff
@@ -0,0 +1,9 @@
+package lksdk
+
+type Private[T any] struct {
+	v T
+}
+
+func MakePrivate[T any](f T) Private[T] {
```
paulwe, line 7: nit: v

paulwe (approved):
lgtm 🚀

pkg/media/pcmlocaltrack.go
```diff
@@ -172,6 +165,10 @@ func (t *PCMLocalTrack) WriteSample(chunk media.PCM16Sample) error {
 		return errors.New("track is closed")
 	}
 
+	if t.muted.Load() {
+		return errors.New("track is muted")
+	}
```
boks1971, line 170: Can we not return an error here? Users could log error if they get an error return. With remote mute, users could still be writing samples and getting an error back. I think it is fine to document that `WriteSample` will drop samples when muted (local or remote) and not return an error here.

pkg/media/pcmlocaltrack.go
```diff
@@ -211,6 +209,26 @@ func (t *PCMLocalTrack) processSamples() {
...
+func (t *PCMLocalTrack) setMuted(muted bool) error {
+	if t.closed.Load() {
+		return errors.New("track is closed")
+	}
+
+	// We will continue to write silence on mute to not mess
+	// up the RTP timestamps. An edge case here is that the mute
+	// value will be checked on every tick (10ms), so any new data on unmute
+	// might be delayed by at most 10ms, which is not much and keeps RTP/RTCP
+	// in sync. The packets will be dropped by the SFU anyway.
```
boks1971, line 221: I am guessing this comment is outdated. The mute check in processSamples() has been removed. You can just mention that on mute, the pending samples are dropped.

anunaym14: Yeah, I removed the mute check because your code would already send silence there, no need to check if track is muted or not. We need to continue to send silence to not mess up the RTP timestamps. So the comment is not outdated. 

So the behavior is: pending samples are dropped and the track writes silence to keep RTP timestamps in check. But these silence packets will be dropped by the SFU anyway.

pkg/media/pcmlocaltrack.go
anunaym14: I'm not too sure about this, they should not be writing samples on mute, so an error makes sense. 

As a user, if I don't have the code to handle remote mute, the error will help me spot it.

boks1971: That means application has to maintain state and not write or filter for mute errors. Yeah, not handling remote mute would be more obvious, but once an app fixes that, it would need to maintain mute state at app level also or filter errors. Don't have strong opinions, but looking a lot of code for specific error handling or maintaining state in two places, wanted to suggest that.

pkg/media/pcmlocaltrack.go
boks1971: The comment says
```
    // An edge case here is that the mute
    // value will be checked on every tick (10ms), so any new data on unmute
    // might be delayed by at most 10ms, which is not much and keeps RTP/RTCP
    // in sync. The packets will be dropped by the SFU anyway.
```

I read that as `processSamples` checks for mute explicitly which it is not doing. And can probably replace 10ms with the constant name (if somebody changes that constant to 20ms, this comment will be outdated). And maybe drop SFU reference, that is behaviour of some remote app which may change at any time. I think the main point here is that this keep sending packets even when muted to maintain RTP timestamps and possibly leave it at that.

anunaym14: Understood, thanks!

PR #684, opened 2025-06-03
Support extra attributes in Join

dennwc:
Adds client-side support for https://github.com/livekit/livekit/pull/3693.


signalclient.go
```diff
@@ -153,6 +155,14 @@ func (c *SignalClient) connectContext(ctx context.Context, urlPrefix string, tok
 			urlSuffix += fmt.Sprintf("&sid=%s", participantSID)
 		}
 	}
+	if len(params.Attributes) != 0 {
+		data, err := json.Marshal(params.Attributes)
+		if err != nil {
+			return nil, ErrInvalidParameter
+		}
+		str := base64.URLEncoding.EncodeToString(data)
```
paulwe, line 163: why are we using base64 instead of url encoding?

davidzhao: it seems more readable versus a bunch of escape chars?

davidzhao (approved):

signalclient.go
paulwe: the server decodes url encoding when parsing the url. base64 requires a separate, manual decoding step.

davidzhao: I think that's true for individual key/vals.. does Go's decoder handle dicts built-in?

paulwe: oh, i wasn't suggesting changing the json - there's no standard for url dict encoding afaik. i was wondering why chose `base64(json(args))` over `urlencode(json(args))` since this requires decoding `json(base64(param))` instead of `json(param)`... base64 is a bit more space efficient but url encoding is the standard query parameter encoding. it's an internal api so we can do whatever we want i just wasn't sure what motivated the decision

dennwc: I was mostly considering the length of the encoding. Although max length for the URL is pretty long, attributes could very long as well. This is why I opted for a more compact encoding.

PR #686, opened 2025-06-04
added LocalTrackSubscribed event for RoomCallbacks

s-hamdananwar:


room.go
```diff
@@ -817,6 +818,15 @@ func (r *Room) handleTranscriptionReceived(transcription *livekit.Transcription)
 	r.callback.OnTranscriptionReceived(transcriptionSegments, p, publication)
 }
 
+func (r *Room) handleLocalTrackSubscribed(trackSubscribed *livekit.TrackSubscribed) {
+	trackPublication := r.LocalParticipant.getLocalPublication(trackSubscribed.TrackSid)
+	if trackPublication == nil {
+		r.log.Debugw("recieved track subscribed for unknown track", "track", trackSubscribed.TrackSid)
```
boks1971: use `trackID` for the track SID key as that is what we use everywhere.

boks1971 (approved):

room.go
s-hamdananwar: @boks1971 The protocol returns `TrackSid (ref [here](https://github.com/livekit/protocol/blob/c8f416a2e6b87f49de12dd6254043fb51a523602/protobufs/livekit_rtc.proto#L462)). should I make a PR to update that as well?

boks1971: No, just meant the logging key. Use `trackID` instead of `track` so that we catch all logs for specific track when searching by that id. Other parts of LiveKit software uses `trackID` as logging key to log the track's SID.

s-hamdananwar: makes sense, thanks!

PR #690, opened 2025-06-12
Support dynacast.

boks1971:


localtrack.go
```diff
@@ -602,6 +602,7 @@ func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
 				}
 			}
 
+			sample.Timestamp = time.Now()
```
boks1971: Use wall clock time in the sample so that gaps in write due to mute would be handled appropriately in `WriteSample`.

publication.go
```diff
@@ -73,7 +73,15 @@ func (p *trackPublicationBase) Track() Track {
...
+			return info.MimeType
+		}
+
+		for _, codec := range info.Codecs {
+			if codec.MimeType != "" {
+				return codec.MimeType
+			}
+		}
 	}
 	return ""
```
boks1971, line 101: This requires some more work. `TrackInfo` has a top level `MimeType` which is not set on server side till the track is published. So, it is not available in the `TrackPublishedResponse`.

But, if client specified `SimulcastCodecs` in `AddTrackRequest`, the `TrackPublishedResponse` will have `Codecs` populated and that will have the `MimeType` specified in `AddTrackRequest`.  Just taking the first one here which has a non-nil `MimeType`. Okay for now as Go SDK does not (yet) support simulcast codec feature.

When it does, this struct needs to be updated to handle multiple mime types. Will actually add this note in code also.

localtrack.go
boks1971: This might need to some more massaging.

Maybe, make a start time and add `Duration` on every new sample to the previous time and then set the `now` time when coming out of mute.

Each of those approaches is going to have some jitter. May need some more experimentation.

room.go
```diff
@@ -843,6 +844,39 @@ func (r *Room) handleLocalTrackSubscribed(trackSubscribed *livekit.TrackSubscrib
...
+			"mime", trackPublication.MimeType(),
+			"subscribedCodec", protoLogger.Proto(subscribedCodec),
+		)
+		for _, subscribedQuality := range subscribedCodec.Qualities {
+			track := trackPublication.GetSimulcastTrack(subscribedQuality.Quality)
+			if track != nil {
+				track.setMuted(!subscribedQuality.Enabled)
+				r.log.Infow(
+					"updating layer enable",
+					"trackID", trackPublication.SID(),
```
anunaym14, line 871: the two info logs look a bit redundant, if we need both, maybe worth it to make one of them debug.

the "updating subscribed quality" log can be a bit misleading if simulcast track does not have the quality (not sure if there is such a case in the e2e flow, just judging by this code)

boks1971: yeah, I was debating what/where to log this. For the first version, thought of logging both the message and also at layer level just to ensure that they are consistent.

Will change the first log to `handling subscribed quality update` for now.

PR #703, opened 2025-07-23
Fix: Correct error condition for handler registration methods

kony428:
This PR fixes the logic for error handling in the following methods:

- RegisterRpcMethod
- RegisterTextStreamHandler
- RegisterByteStreamHandler

Previously, these methods return an error even when the handler was not actually registered, due to incorrect error condition checks. The logic is now corrected so that an error is only returned if a handler for the given method or topic is already registered.


anunaym14 (approved):
Good catch! 

My bad on this one, might have mixed it up with something else. Anyway, thanks for the PR!

Please sign the CLA, so that I can go ahead and merge this.

PR #706, opened 2025-07-27
Get to the point of establishing publishing peer connection and opening signalling data channel.

boks1971:
- Adding a signalling version option to avoid changing code.

Need to do
- establishing signalling data channel should be driven by signal transport and not signalling version. For now, it is based on signalling version.
- have to configure publisher PC at start up and use it in `ConnectRequest`. That requires more wrangling of code which I will do later.

### DIFF
--- a/signalling/signaltransport_datachannel.go
+++ b/signalling/signaltransport_datachannel.go
@@ -15,12 +15,8 @@
 package signalling
 
 import (
-	"math/rand"
-	"sync"
-
 	"github.com/livekit/protocol/logger"
 	"github.com/pion/webrtc/v4"
-	"go.uber.org/atomic"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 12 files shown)


signalling/signaltransport_datachannel.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	if !s.isStarted {
+		return ErrTransportNotStarted
+	}
+
+	select {
+	case s.msgChan <- msg:
```
boks1971: Need to consolidate this message queue as this and signaltransport_http use similar bits.

boks1971: Created the message queue in this [commit](https://github.com/livekit/server-sdk-go/pull/706/commits/2395d502a03a882555488a549b04a86ad3f171fc).

paulwe (approved):

PR #709, opened 2025-08-01
Establish publisher PC along with connect in v2.

boks1971:


engine.go
```diff
@@ -520,6 +486,100 @@ func (e *RTCEngine) configure(
...
+		if pair, err := transport.GetSelectedCandidatePair(); err == nil {
+			fields = append(fields, "transport", signalTarget, "iceCandidatePair", pair)
+		}
+		e.log.Debugw("ICE connected", fields...)
+	case webrtc.ICEConnectionStateDisconnected:
+		e.log.Debugw("ICE disconnected", "transport", signalTarget)
+	case webrtc.ICEConnectionStateFailed:
+		e.log.Debugw("ICE failed", "transport", signalTarget)
+		if isPrimary {
+			e.handleDisconnect(false)
```
cnderrauber, line 563: should we handle non-primary disconnect as well?

cnderrauber (approved):

engine.go
boks1971: Sorry, missed this before merging @cnderrauber . Yeah, good point. No reason to not handle it. I just repeated what was there. Will change it in next commit to handle it for both.

PR #710, opened 2025-08-04
Handle disconnect of non-primary and log first response time.

boks1971:


engine.go
```diff
@@ -559,9 +564,7 @@ func (e *RTCEngine) handleICEConnectionStateChange(
 		e.log.Debugw("ICE disconnected", "transport", signalTarget)
 	case webrtc.ICEConnectionStateFailed:
 		e.log.Debugw("ICE failed", "transport", signalTarget)
-		if isPrimary {
-			e.handleDisconnect(false)
-		}
+		e.handleDisconnect(false)
```
boks1971, line 567: @cnderrauber Addressing comment from previous PR. Handling disconnect on non-primary PC also.

PR #711, opened 2025-08-06
Delete v2 signalling

boks1971:


engine.go
```diff
@@ -248,35 +219,19 @@ func (e *RTCEngine) JoinContext(
...
-		publisherOffer, err = e.publisher.GetOffer()
-		if err != nil {
-			e.pclock.Unlock()
-			return false, err
-		}
-		e.pclock.Unlock()
-	}
-
-	err = e.signalTransport.Join(ctx, url, token, *connectParams, publisherOffer)
-	if err != nil {
```
boks1971, line 268: Will bring back sending publisher offer in the first message in a separate PR later.

cnderrauber (approved):

PR #723, opened 2025-08-25
Fixing offset calculation

milos-lk:
Sender reports shouldn't overwrite the desired offset but update it. Also signal the correct diff in onSR callbacks.


frostbyte73 (approved):

frostbyte73:
> Don't know how exactly the synchronizer works, so would be good to get more 👀 on this, but wanted to ask if this solves the track disappearing issue?

Yes, if the track is published more than 2s after the egress was started. We had such a hard time finding a reproduction because all of our integration tests publish before starting the egress

PR #724, opened 2025-08-26
Single peer connection support

boks1971:
Conditioned on version  >= 3.0.0

Tested by manually setting version to 3.0.0 and doing the publish demo using lk cli.


engine.go
```diff
@@ -445,6 +448,10 @@ func (e *RTCEngine) createPublisherPCLocked(configuration webrtc.Configuration)
 }
 
 func (e *RTCEngine) createSubscriberPCLocked(configuration webrtc.Configuration) error {
+	if semver.Compare("v"+Version, "v3.0.0") >= 0 {
+		return nil
+	}
```
anunaym14: The semver comparison is done in two different struct methods, we should probably add a field `singePCMode` or something and set it while creating the engine

localparticipant.go
```diff
@@ -67,8 +67,8 @@ func (p *LocalParticipant) PublishTrack(track webrtc.TrackLocal, opts *TrackPubl
 		}
 	}
 
-	publisher, ok := p.engine.Publisher()
-	if !ok {
+	transport := p.getPublishTransport()
+	if transport == nil {
 		return nil, ErrNoPeerConnection
 	}
```
anunaym14, line 73: The lack of transport should be handled internally right?

room.go
```diff
@@ -552,13 +572,16 @@ func (r *Room) sendSyncState() {
...
 		Subscription: &livekit.UpdateSubscription{
 			TrackSids: trackSids,
 			Subscribe: !sendUnsub,
 		},
-		PublishTracks: publishedTracks,
-		DataChannels:  dataChannels,
+		PublishTracks:     publishedTracks,
+		DataChannels:      dataChannels,
+		TrackSidsDisabled: trackSidsDisabled,
+		// MIGRATION-TODO DatachannelReceiveStates
```
anunaym14, line 584: Pending TODO, just wanted to highlight it

engine.go
anunaym14: The same comparison is done in Room too, we can do it there and pass it to the engine

room.go
boks1971: Yeah, I added it for attention later.

localparticipant.go
boks1971: Probably. But, I am not changing behaviour here.

engine.go
boks1971: Will do.

localparticipant.go
anunaym14: Yeah, it's not from this PR, but something we can make a note of for v3

PR #731, opened 2025-08-29
Fix handling of sender report offsets

milos-lk:
While the recent change made sure packet PTSs don't jump unexpectedly, it also added wrong cumulative handling of offsets.
Desired PTS offset needs to be split in 2 components:
- basedPTS offset (basically offset from the first track appearance) which needs to be kept constant
- offset introduced on sender reports - which shouldn't be accumulated but overwritten

In addition to the change adding some synchronizer unit tests to cover  basic PTS generation logic.


go.mod
```diff
@@ -87,3 +92,5 @@ require (
 	gopkg.in/hraban/opus.v2 v2.0.0-20230925203106-0188a62cb302 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+tool github.com/maxbrunsfeld/counterfeiter/v6
```
boks1971, line 96: is this required? Guess it makes things repeatable/consistent?

pkg/synchronizer/track.go
```diff
@@ -59,6 +62,7 @@ type TrackSynchronizer struct {
 	// offsets
 	currentPTSOffset time.Duration // presentation timestamp offset (used for a/v sync)
 	desiredPTSOffset time.Duration // desired presentation timestamp offset (used for a/v sync)
+	basePTSOffset    time.Duration // component of the desired PTS offset (set initially to preseve initial offset)
```
boks1971: :nit: `preserve` spelling

pkg/synchronizer/track.go
```diff
@@ -171,7 +176,7 @@ func (t *TrackSynchronizer) onSenderReport(pkt *rtcp.SenderReport) {
 		return
 	}
 
-	t.desiredPTSOffset += offset
```
boks1971, line 174: Is there a chance that PTS can move backward because of this?

pkg/synchronizer/synchronizer_test.go
```diff
@@ -0,0 +1,254 @@
...
+	tr := fakeAudio48k(0xA0010001)
+
+	ts := uint32(1_000_000)
+	tsync := s.AddTrack(tr, "p1")
+
+	tsync.Initialize(pkt(ts))
+
+	adj0, err := tsync.GetPTS(pkt(ts))
+	require.NoError(t, err)
+	near(t, adj0, 0, 5*time.Millisecond)
```
boks1971: is 5ms enough? some of our GH runners are super slow. Ideally, using a mock clock would be great, but that is not easy to do.

boks1971 (approved):
lgtm!

pkg/synchronizer/track.go
milos-lk: yes - it could keep accumulating in both directions depending on SRs. The version before was simply overwriting `t.desiredPTSOffset` which obviously breaks track starting late.

go.mod
milos-lk: go get -tool github.com/maxbrunsfeld/counterfeiter/v6
is what it got it added there - I think it's fine to have it available there for go tool for this module.

pkg/synchronizer/synchronizer_test.go
milos-lk: this is good point - most of variation comes from potentially variable time needed between setting t.startTime and calculating estimatedPTS few instructions later (don't expect huge diff there unless the gh runner is super super slow). I could bump the tolerance a bit higher and keep an eye on this - and if flakiness appear  I could introduce mock clock (which due to its invasiveness would like to skip if we can get away without it). wdyt?

boks1971: yeah, good strategy, mock clock will require a lot of changes. 

Most of the time it should be fine. Once in a while, runner is super slow and probably have some noisy neighbour hogging cpu and causing flakiness, but it should be very rare.

frostbyte73 (approved):

PR #739, opened 2025-09-25
Log start time and drop audio in pts disable config.

boks1971:
PTS disable is not audio specific. So, changing it to just call it PTSAdjustmentDisabled. Changes the API a bit, but I think it is the correct usage as nothing in track synchroniser that is audio specific.

### DIFF
--- a/pkg/synchronizer/track.go
+++ b/pkg/synchronizer/track.go
@@ -49,8 +49,8 @@ type TrackSynchronizer struct {
 	*rtpConverter
 
 	// config
-	maxTsDiff                   time.Duration // maximum acceptable difference between RTP packets
-	audioPTSAdjustmentsDisabled bool          // disable audio packets PTS adjustments on SRs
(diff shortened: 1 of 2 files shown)


pkg/synchronizer/track.go
```diff
@@ -131,8 +137,9 @@ func (t *TrackSynchronizer) GetPTS(pkt *rtp.Packet) (time.Duration, error) {
 	estimatedPTS := time.Since(t.startTime)
 	if pts < t.lastPTS || !t.acceptable(pts-estimatedPTS) {
 		newStartRTP := ts - t.toRTP(estimatedPTS)
-		t.logger.Infow("correcting PTS",
-			"TS", ts,
+		t.logger.Infow(
+			"correcting PTS",
+			"currentTS", ts,
```
boks1971, line 142: changing this as `ts` is a reserved field in DataDog. This particular case was okay as it is case sensitive and this still got displayed, but don't want to risk collision.

PR #741, opened 2025-09-25
Use a one-way-delay estimator to estimate propagation delay.

boks1971:
This PR adds a way to estimate one-way-delay and use it for adjustments to PTS in synchronizer.

There are two options now
1. Existing method
2. Using one-way-delay estimator
Default is the existing method

Changes to existing method
- Use arrival time of packet, i. e. when the packet was read off the wire and pushed into jitter buffer as start time of synchronizer and drop old packets.
   - Made two configs for these, one to control start time and one to set a threshold for old packet drops. Made them two to make it explicit. But, can make it one, i. e. if the old packet threshold is non-zero, can switch to using packet arrival time.
- Limit the amount of PTS adjustment in a time window. Setting it to max 2ms in a 100ms window.

One-way-delay method
- Use receive time of packet for start time. This allows packets pulled late from jitter buffer to have a good start time or be dropped to keep alignment.
- Use one-way-delay estimator to estimate propagation delay between sender and receiver. And rebase sender NTP time using that so that operations are in local clock domain.
- Based on rebased sender report, adjust start time for the first two minutes to get a good estimate of start time. This will make subsequent PTS calculations more reliable.
- Ensure that PTS is always moving forward.
- If this option is enabled, `audioPTSAdjustmentsDisabled` config is ignored, i. e. PTS adjustment is always applied, rationale being the existing method should be used if disabling the adjustments.

Will leave more notes inline.


pkg/synchronizer/track.go
```diff
@@ -206,6 +245,80 @@ func (t *TrackSynchronizer) onSenderReport(pkt *rtcp.SenderReport) {
...
+		receivedAt,
+	)
+	if stepChange {
+		t.logger.Debugw(
+			"propagation delay step change",
+			"propagationDelayEstimator", t.propagationDelayEstimator,
+		)
+	}
+
+	return time.Duration(estimatedPropagationDelay)
```
boks1971: there is some back and forth between `int64` and `time.Duration`. Would really like to change everything to `int64` so that it is easier on my brain :-) and convert to `time.Duration` only when necessary.

pkg/synchronizer/track.go
```diff
@@ -160,6 +169,24 @@ func (t *TrackSynchronizer) GetPTS(pkt *rtp.Packet) (time.Duration, error) {
...
+			"lastPTS", t.lastPTS,
+			"estimatedPTS", estimatedPTS,
+			"offset", pts-estimatedPTS,
+			"startRTP", t.startRTP,
+			"propagationDelay", t.propagationDelayEstimator,
+			"totalStartTimeAdjustment", t.totalStartTimeAdjustment,
+			"adjustedPTS", adjusted,
+			"lastPTSAdjusted", t.lastPTSAdjusted,
+		)
+		adjusted = t.lastPTSAdjusted + time.Millisecond
```
boks1971, line 413: probably needs some tweaking to see how far off the offset is and how much to adjust. I don't have a good feel for how this looks without running some experiments.

pkg/synchronizer/track.go
```diff
@@ -179,6 +206,10 @@ func (t *TrackSynchronizer) onSenderReport(pkt *rtcp.SenderReport) {
 	t.Lock()
 	defer t.Unlock()
 
+	// estimate propagation, i. e. one way delay based on NTP time in the report and when it is received
+	receivedAt := mono.UnixNano()
+	estimatedPropagationDelay := t.updatePropagationDelay(pkt, receivedAt)
+
 	if pkt.RTPTime == t.lastSR || t.startTime.IsZero() {
```
boks1971: would be good to make these 64-bit so that we can do checks for out-of-order easily. Even here, I think this should do a out-of-order check and drop.

pkg/synchronizer/track.go
milos-lk: this will always land in the pervious buffer range [t.lastPTSAdjusted + previous packet_duration (e.g 20ms)), not sure if it's fine to just leave it like that - this packet will probably be dropped.

pkg/synchronizer/owd_estimator.go
```diff
@@ -0,0 +1,179 @@
+// Copyright 2023 LiveKit, Inc.
```
boks1971, line 1: This currently is also in OSS livekit repo. Can move it to some central location if this is adopted here also.

pkg/synchronizer/track.go
```diff
@@ -133,33 +147,90 @@ func (t *TrackSynchronizer) GetPTS(pkt *rtp.Packet) (time.Duration, error) {
 		return t.lastPTSAdjusted, nil
 	}
 
+	if (ts - t.lastTS) > (1 << 31) {
+		t.logger.Infow(
+			"dropping out-of-order packet",
+			"currentTS", ts,
+			"lastTS", t.lastTS,
+		)
+		return 0, errors.New("out-of-order")
```
milos-lk: this effectively ends recording (for the track) here
https://github.com/livekit/egress/blob/main/pkg/pipeline/source/sdk/appwriter.go#L313

I guess we explicitly return error here just for debugging?

boks1971: Oh, good to know. Intention is to drop it. Is there a way to drop a packet?

milos-lk: as far as I can see we need to handle the specific error at caller site - there we could drop it if error is of the type

pkg/media/pcmremotetrack.go
```diff
@@ -158,13 +158,14 @@ func (t *PCMRemoteTrack) processSamples(handleJitter bool) {
 		h = newDecryptionHandler(h, t.decryptor)
 	}
 
+	hc := newHandlerCloser(h)
```
boks1971, line 161: unrelated to the functionality of this PR, but pulling in the latest media-sdk necessitated this change.

pkg/synchronizer/synchronizer.go
```diff
@@ -22,16 +22,20 @@ import (
 )
 
 const (
-	DefaultMaxTsDiff = time.Minute
+	DefaultMaxTsDiff          = time.Minute
+	DefaultOldPacketThreshold = 500 * time.Millisecond
```
boks1971: Using a default of 0.5 second for old packets, probably needs tweaking. Can pass in a config to modify this though.

pkg/synchronizer/track.go
```diff
@@ -115,24 +148,62 @@ func (t *TrackSynchronizer) Initialize(pkt *rtp.Packet) {
...
+	}
+
+	lastTS := t.lastTS
+	if lastTS == 0 {
+		lastTS = ts
+	}
+
 	pts := t.lastPTS + t.toDuration(ts-t.lastTS)
 	estimatedPTS := time.Since(t.startTime)
 	if pts < t.lastPTS || !t.acceptable(pts-estimatedPTS) {
```
milos-lk, line 207: is it possible that if reception time is used this check becomes too sensitive to jitter.Buffer latency?

boks1971: Yeah good point. I am thinking if we should drop this check entirely.

milos-lk: yeah - I guess with discarding too old buffer we are safe wrt lower bound of that acceptable range - right? crossing upper bound (too far in future) might still be an issue.

milos-lk: just thinking about it - if we receive a burst of packets relatively fast - this check won't protect us also wrt upper bound in the way it was designed.

pkg/synchronizer/track.go
```diff
@@ -16,19 +16,36 @@ package synchronizer
...
+	// Throttle PTS adjustment to a limited amount ina time window.
+	// This setting determines how long a certain amount of adjustment
+	// throttles the next adjustment.
+	//
+	// For sample, if a 1ms adjustment is appied at 1%, it means that
+	// 1ms is 1% of ajustment window, so the adjustement window is 100ms
+	// and next adjustment will not applied for till tha time elapses
+	//
+	// With the settings of 2ms adjustment at 2%, a mamximum adjustment
+	// of 2ms per 100ms
```
milos-lk: just a note for me to make sure this one lower than min adjustment step:
https://github.com/livekit/egress/blob/add52340e270e784cf8706989453afd4d0791a1b/pkg/pipeline/builder/audio.go#L42

pkg/synchronizer/track.go
boks1971: True as long as the old packet threshold than maxTsDiff I think.

Also, another thing I was thinking about is to drop old packets only at the start and not do it once the stream starts. What do you think about that?

milos-lk: that's also an option - the question remains how do we handle cases when too old packets appear midstream - if we want to play catch up game (as we do so far) - we rebase them to the current window and by doing that we bring the track back making sure it doesn't disappear (or cause others to disappear at mixer level) but at the expense of broken sync between tracks.. I am thinking - maybe it's better to drop these packets in a hope that the source will catch up - so essentially we trade a gap in the signal for keeping tracks sync - and we could maybe rely on PLC logic to cover the gap if small enough. Question is what if the source doesn't "catch up" - in which case we would have a track which is preeminently behind the current acceptable window and I am not sure if there is anything we can do to bring it "back" without affecting the inter track sync.

milos-lk: this also brings the case of what to do when track goes ahead... at the moment we pull it back and essentially do the same as described above and break inter track sync. Maybe it's better to pace it (and let packets be dropped if max channel capacity is reached). Essentially dropping packets again :). Pacing isn't needed but could reduce the number of dropped buffers in case of a surge of packets in a short time window.

boks1971: yeah, two sort of conflicting requirements
1. Keep as much of the stream as possible without dropping packets 
2. Keep sync between tracks when multiple tracks are involved.
Have to decide what is more important. Maybe, there is a mode that can be given to synchronizer to control for that.

pkg/synchronizer/track.go
```diff
@@ -141,25 +217,202 @@ func (t *TrackSynchronizer) GetPTS(pkt *rtp.Packet) (time.Duration, error) {
...
 
 	if t.shouldAdjustPTS() {
+		prevCurrentPTSOffset := t.currentPTSOffset
+		if t.currentPTSOffset > t.desiredPTSOffset {
+			t.currentPTSOffset = max(t.currentPTSOffset-cMaxAdjustment, t.desiredPTSOffset)
+		} else if t.currentPTSOffset < t.desiredPTSOffset {
+			t.currentPTSOffset = min(t.currentPTSOffset+cMaxAdjustment, t.desiredPTSOffset)
+		}
+
+		// throttle further adjustment till a window proportional to this adjustment elapses
```
milos-lk, line 236: if we don't change something inside gst pipeline - the throttling is going to create more audible artifacts. Do you think it would be safer not to throttle in this get variant - to preserve the default behavior?

boks1971: I am thinking in conversational audio, smaller adjustments might better? But, for music and test tones, it will be bad. Can make this a config also so that it can be controlled.

boks1971: A counter argument would be "why limit it to 5ms max and why not apply the entire thing at once? Why apply 5ms per sample?". Anyhow, I am making configs for these.

milos-lk: that is indeed a good question, I was also thinking about that. If we have a negative offset - let's say 120ms - if we apply it at once - the resulting timeline coming out GetPTS call will jump back by 120ms - and all packets worth 120ms coming after the SR will be discarded. So we need to pick a poison :)
Not sure maybe @frostbyte73 has some more context on the existing logic?

boks1971: Parameterised in this [commit](https://github.com/livekit/server-sdk-go/pull/741/commits/1904b2890791802c401d8201059cf92d9b7fd783).

boks1971:
Thank you for the review @milos-lk . I am going to merge this. If there is more feedback, will address in a subsequent PR. This is getting long already and does not change default/existing behaviour.

pkg/synchronizer/track.go
```diff
@@ -206,25 +467,193 @@ func (t *TrackSynchronizer) onSenderReport(pkt *rtcp.SenderReport) {
...
+			"sr", sr,
+			"timeSinceReceive", timeSinceReceive,
+			"timeSinceStart", timeSinceStart,
+			"samplesDiff", samplesDiff,
+			"samplesDuration", samplesDuration,
+		}
+	}
+
+	if adjustedStartTimeNano < startTimeNano {
+		if startTimeNano-adjustedStartTimeNano > cStartTimeAdjustThreshold.Nanoseconds() {
```
milos-lk, line 611: was thinking about this one some more - so far for small time corrections we have done it in small steps of 5ms. That plays nicely with audiorate element further down in the gst pipeline. Its tolerance is currently set to 3ms. This is nice because if the diff between the "perfect" and gotten PTS on the audiorate input is < tolerance it will stitch new buffer at the "perfect" PTS (something we didn't have). That could overtime cause buildup of the these hidden micro jumps. This particular phase lasts for 2 minutes only - but I think it might be safer to perform time manupulation using stair steps changes with delta >  audiorate tolerance.

boks1971: Was trying to do this only for the first few minutes, the idea being it should lock to a pretty good start point with the small adjustments needed in that time.

Do you think we should just get rid of start time adjustment? That should be fine with the adjustments happening later. The start time time adjustment is an attempt to pull things in sync better over time.

PR #743, opened 2025-09-29
Log failure to keep up with real time from the write worker

milos-lk:


localtrack.go
```diff
@@ -613,6 +613,7 @@ func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
 		nextSampleTime = nextSampleTime.Add(sample.Duration)
 		sleepDuration := time.Until(nextSampleTime)
 		if sleepDuration <= 0 {
+			s.log.Warnw("sample provider too slow, can't keep up with real time", nil, "delta", sleepDuration)
```
boks1971, line 616: Unfortunately, this is not always the case @milos-lk . There are things like (I believe SPS/PPS in H.264) which have 0 sample duration and it hits this case.

milos-lk: good point - closing it

PR #744, opened 2025-09-29
Handle 0 duration samples.

boks1971:
With H.264, there could be 0 duration samples, for example SPS/PPS NALUs. They should not increment the RTP timestamp.

Should also pick up https://github.com/pion/interceptor/pull/370 for slightly more accurate sender reports.


localtrack.go
```diff
@@ -579,7 +584,8 @@ func (s *LocalTrack) writeWorker(provider SampleProvider, onComplete func()) {
 	audioProvider, isAudioProvider := provider.(AudioSampleProvider)
 
 	nextSampleTime := time.Now()
-	ticker := time.NewTicker(10 * time.Millisecond)
+
+	ticker := time.NewTicker(time.Hour)
```
boks1971, line 588: Just changed the initial tick to some really large number so that there is no chance of something happening before the first sample.

PR #752, opened 2025-10-02
Set initial PTS to estimated PTS to absorb pipeline start latency.

boks1971:


anunaym14 (approved):
Build failure, but lgtm otherwise!

milos-lk (approved):

PR #759, opened 2025-10-06
Adding burst estimator based start gate for initializing track synchronizer

milos-lk:
RTP packets arriving at a receiver site might come in a burst because of reconnect sending older packets through a NACK + retransmission. Anchoring track start time during phases of bursts would result in setting wrong start time for the track and breaking inter-track synchronization. Adding burst esitmator gate before init which will check between packets RTP and reception time and would allow init only when pace stabilizes (no burst is ongoing).

### DIFF
--- a/pkg/synchronizer/start_gate.go
+++ b/pkg/synchronizer/start_gate.go
@@ -1,0 +1,161 @@
+package synchronizer
+
+import (
+	"time"
+
+	"github.com/livekit/media-sdk/jitter"
+	"github.com/pion/webrtc/v4"
+)
+
+// startGate is a lightweight buffer that decides when a track should be
+// considered "live". Callers push packets into the gate until it returns a
+// slice of packets that can be used to initialize downstream synchronisation.
+// The second return value exposes how many packets were discarded while the
+// gate was waiting for stability, and the third return value indicates whether
+// the gate has finished its job.
+type startGate interface {
+	Push(pkt jitter.ExtPacket) ([]jitter.ExtPacket, int, bool)
+}
+
+// burstEstimatorGate implements startGate using the burst-estimation logic.
+// It buffers packets until their arrival cadence matches
+// the RTP timestamp spacing well enough to assume we are caught up with the
+// realtime stream. The gate can be configured with different tolerances for
+// audio and video since video keyframes may span hundreds of packets.
+type burstEstimatorGate struct {
+	clockRate   uint32
+	maxSkew     time.Duration
+	minArrival  time.Duration
+	scoreTarget int
+
+	score       int
+	lastTS      uint32
+	lastArrival time.Time
+	hasLast     bool
+	done        bool
+	buffer      []jitter.ExtPacket
+	maxBuffer   int
+}
+
+func newStartGate(clockRate uint32, kind webrtc.RTPCodecType) startGate {
+	be := &burstEstimatorGate{
+		clockRate:   clockRate,
+		maxSkew:     20 * time.Millisecond,
+		minArrival:  2 * time.Millisecond,
+		scoreTarget: 5,
+		maxBuffer:   1000, // high bitrate key frames can span hundreds of packets
+	}
+
+	if kind == webrtc.RTPCodecTypeAudio {
+		be.maxSkew = 8 * time.Millisecond
+		be.maxBuffer = 200
+	}
+
+	return be
+}
+
+// Push feeds one packet into the gate. Once pacing stabilizes it returns the
+// buffered packets that should be used to initialize the track synchronizer,
+// along with the number of packets dropped while waiting and a done flag.
+func (b *burstEstimatorGate) Push(pkt jitter.ExtPacket) ([]jitter.ExtPacket, int, bool) {
+	if b.done {
+		return nil, 0, true
+	}
+
+	if len(pkt.Payload) == 0 {
+		return nil, 0, false
+	}
+
+	if !b.hasLast {
+		b.lastTS = pkt.Timestamp
+		b.lastArrival = pkt.ReceivedAt
+		b.hasLast = true
+		b.buffer = append(b.buffer[:0], pkt)
+		return nil, 0, false
+	}
+
+	signedTsDelta := int32(pkt.Timestamp - b.lastTS)
+	arrivalDelta := pkt.ReceivedAt.Sub(b.lastArrival)
+
+	if signedTsDelta < 0 {
+		// ignore out-of-order packets while continuing to wait for stability
+		return nil, 0, false
+	}
+
+	if signedTsDelta == 0 {
+		// multiple packets with the same timestamp (e.g. key frame)
+		b.buffer = append(b.buffer, pkt)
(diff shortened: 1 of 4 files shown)


pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+	return time.Duration(int64(delta) * int64(time.Second) / int64(b.clockRate))
+}
+
+func (b *burstEstimatorGate) enforceWindow() int {
+	if b.maxBuffer == 0 || len(b.buffer) <= b.maxBuffer {
+		return 0
+	}
+	dropped := len(b.buffer) - b.maxBuffer
+	copy(b.buffer, b.buffer[dropped:])
+	b.buffer = b.buffer[:len(b.buffer)-dropped]
```
boks1971: sorry I am having difficulty understanding this, what is the copy and re-init doing.

Assuming max is 100 and there are 120 packets in there.
- dropped = 20
- packets 20 - 119 get copied to  0 - 99
- new buffer is pointing to 0 - 99

Would something like `b.buffer = b.buffer[dropped:]` achieve the same thing?

pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil, dropped, false
+		}
+		return nil, 0, false
+	}
+
+	b.lastTS = pkt.Timestamp
+	b.lastArrival = pkt.ReceivedAt
+
+	tsDuration := b.timestampToDuration(uint32(signedTsDelta))
+	if tsDuration <= 0 {
```
boks1971: this condition should not happen right with the checks ☝️ ? Guess it can be 0 if clock rate is 0, but that is just not right config

pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+	done        bool
+	buffer      []jitter.ExtPacket
+	maxBuffer   int
+}
+
+func newStartGate(clockRate uint32, kind webrtc.RTPCodecType) startGate {
+	be := &burstEstimatorGate{
+		clockRate:   clockRate,
+		maxSkew:     20 * time.Millisecond,
+		minArrival:  2 * time.Millisecond,
```
boks1971: can we make this a percentage of expected diff? For example, it two packets are to be separated by 20ms, can we say anything less than 20% of that number is a burst. Always a little skeptical of absolute numbers :-)

pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+	if tsDuration <= 0 {
+		dropped := len(b.buffer)
+		b.buffer = b.buffer[:0]
+		b.score = 0
+		b.buffer = append(b.buffer, pkt)
+		return nil, dropped, false
+	}
+
+	if arrivalDelta < b.minArrival {
+		dropped := len(b.buffer)
```
boks1971: For this kind of code block, I usually make a small local function variable and call it to avoid repeated code like this. Works better if I have to change something later. Just have to change one place.

pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+		b.buffer = b.buffer[:0]
+		b.score = 0
+		b.buffer = append(b.buffer, pkt)
+		return nil, dropped, false
+	}
+
+	skew := arrivalDelta - tsDuration
```
boks1971, line 130: Made a note about making `minArrival` a percentage of expected duration. Can we do the same here? i. e. if the skew is like 150% (or maybe even 200%) of the expected, treat it as a bad sample.

pkg/synchronizer/start_gate.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, dropped, false
+	}
+
+	b.buffer = append(b.buffer, pkt)
+	if dropped := b.enforceWindow(); dropped > 0 {
+		return nil, dropped, false
+	}
+
+	if b.score < b.scoreTarget {
+		b.score++
```
boks1971: Can we do a hybrid check here?
Number of packets or duration of good sequence. With DTX, there will be a packet every 400 ms. So, something like a certain number of packets or 1s whichever happens earlier can be considered a good sequence.

boks1971 (commented):
lgtm! @milos-lk , left a few suggestions, please let me know what you think.

pkg/synchronizer/start_gate.go
milos-lk: yeah - was too defensive here - will ditch it

pkg/synchronizer/start_gate.go
milos-lk: let me check  - this should be possible

pkg/synchronizer/start_gate.go
milos-lk: functionally same thing - I think I went too far ahead trying to over optimize this - the only diff is that `b.buffer = b.buffer[dropped:]` would reduce the slice capacity so we might have more memory allocations - but since this is done only once I think I will just re-slice it.

boks1971:
One other suggestion @milos-lk . Can we log the amount of time it takes to resolve? Wondering if we should also have a safety net and just init if the time elapsed for the resolve exceeds something like 2s?

Basically log things like score, time since start, time of good sequence so that we get some data on how the gate works.

pkg/synchronizer/start_gate.go
milos-lk: updated

pkg/synchronizer/start_gate.go
milos-lk: removed

pkg/synchronizer/start_gate.go
milos-lk: updated

pkg/synchronizer/start_gate.go
milos-lk: updated

pkg/synchronizer/start_gate.go
milos-lk: updated

pkg/synchronizer/start_gate.go
milos-lk: updated

boks1971 (approved):
🚀

pkg/synchronizer/track.go
```diff
@@ -139,6 +139,10 @@ func (t *TrackSynchronizer) OnSenderReport(f func(drift time.Duration)) {
 func (t *TrackSynchronizer) PrimeForStart(pkt jitter.ExtPacket) ([]jitter.ExtPacket, int, bool) {
 	if t.initialized || t.startGate == nil {
 		if !t.initialized {
+			// skip dummy packets
+			if len(pkt.Payload) == 0 {
+				return nil, 0, false
+			}
```
boks1971: I am thinking this can be dropped even if `startGate != nil`.

I think `startGate` will drop it based on same time stamp, but I think it will be better to drop here explicitly till initialised irrespective of whether start gate is there or not.

PR #765, opened 2025-10-10
Revive track that constantly falls behind the current timeline

milos-lk:
We have seen cases where a track RTP packets timestamps could fall behind live mixer thereshold but still be in the acceptable range. Lowering acceptable range is avoided here to be able to avoid time rebase for the cases where a track only temporarly falls behind and then catches up - we would still like to avoid rebasing in that case so that we don't introduce syncrhonization problems (few missed packets at the cost of preserved sync).
The idea of this change is that if a track is constantly behind for some time - it's better to pull it forward in time and prevent it from being completely lost (at the cost of skewed synchronizaiton)

### DIFF
--- a/pkg/synchronizer/track.go
+++ b/pkg/synchronizer/track.go
@@ -69,16 +69,18 @@ type TrackSynchronizer struct {
 	enableStartGate                   bool
 
 	// timing info
-	startTime        time.Time // time at initialization --> this should be when first packet is received
-	startRTP         uint32    // RTP timestamp of first packet
-	firstTime        time.Time // time at which first packet was pushed
-	lastTS           uint32    // previous RTP timestamp
-	lastSN           uint16    // previous RTP sequence number
-	lastTime         time.Time
-	lastPTS          time.Duration // previous presentation timestamp
-	lastPTSAdjusted  time.Duration // previous adjusted presentation timestamp
-	lastTSOldDropped uint32        // previous dropped RTP timestamp due to old packet
-	maxPTS           time.Duration // maximum valid PTS (set after EOS)
+	startTime              time.Time // time at initialization --> this should be when first packet is received
+	startRTP               uint32    // RTP timestamp of first packet
+	firstTime              time.Time // time at which first packet was pushed
+	lastTS                 uint32    // previous RTP timestamp
+	lastSN                 uint16    // previous RTP sequence number
+	lastTime               time.Time
+	lastPTS                time.Duration // previous presentation timestamp
+	lastPTSAdjusted        time.Duration // previous adjusted presentation timestamp
+	lastTSOldDropped       uint32        // previous dropped RTP timestamp due to old packet
+	maxPTS                 time.Duration // maximum valid PTS (set after EOS)
+	oldSequenceDuration    time.Duration // duration the sequence of of old and discarded packets
+	fallingBehindThreshold time.Duration // threshold at which a track is considered falling behind
 
 	// offsets
 	currentPTSOffset           time.Duration // presentation timestamp offset (used for a/v sync)
@@ -114,6 +116,7 @@ func newTrackSynchronizer(s *Synchronizer, track TrackRemote) *TrackSynchronizer
 		preJitterBufferReceiveTimeEnabled: s.config.PreJitterBufferReceiveTimeEnabled,
 		rtcpSenderReportRebaseEnabled:     s.config.RTCPSenderReportRebaseEnabled,
 		oldPacketThreshold:                s.config.OldPacketThreshold,
+		fallingBehindThreshold:            s.config.FallingBehindThreshold,
 		enableStartGate:                   s.config.EnableStartGate,
 		nextPTSAdjustmentAt:               mono.Now(),
 		propagationDelayEstimator:         NewOWDEstimator(OWDEstimatorParamsDefault),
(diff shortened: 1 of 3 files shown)


pkg/synchronizer/track.go
```diff
@@ -69,16 +69,18 @@ type TrackSynchronizer struct {
...
+	startRTP               uint32    // RTP timestamp of first packet
+	firstTime              time.Time // time at which first packet was pushed
+	lastTS                 uint32    // previous RTP timestamp
+	lastSN                 uint16    // previous RTP sequence number
+	lastTime               time.Time
+	lastPTS                time.Duration // previous presentation timestamp
+	lastPTSAdjusted        time.Duration // previous adjusted presentation timestamp
+	lastTSOldDropped       uint32        // previous dropped RTP timestamp due to old packet
+	maxPTS                 time.Duration // maximum valid PTS (set after EOS)
+	oldSequenceDuration    time.Duration // duration the sequence of of old and discarded packets
```
boks1971, line 82: :nit: two of

pkg/synchronizer/track.go
```diff
@@ -114,6 +116,7 @@ func newTrackSynchronizer(s *Synchronizer, track TrackRemote) *TrackSynchronizer
 		preJitterBufferReceiveTimeEnabled: s.config.PreJitterBufferReceiveTimeEnabled,
 		rtcpSenderReportRebaseEnabled:     s.config.RTCPSenderReportRebaseEnabled,
 		oldPacketThreshold:                s.config.OldPacketThreshold,
+		fallingBehindThreshold:            s.config.FallingBehindThreshold,
```
boks1971, line 119: log this in the `initialized track synchronizer` log line so that we know all the config settings.

pkg/synchronizer/track.go
boks1971, line 82: log this in `MarshalLogObject` so that we know the state always

pkg/synchronizer/track.go
```diff
@@ -774,6 +795,10 @@ func (t *TrackSynchronizer) isPacketTooOld(packetTime time.Time) bool {
 	return t.oldPacketThreshold != 0 && mono.Now().Sub(packetTime) > t.oldPacketThreshold
 }
 
+func (t *TrackSynchronizer) trackFallingBehind() bool {
+	return t.oldSequenceDuration > t.fallingBehindThreshold
```
boks1971, line 799: Should 0 for threshold mean disabled? i. e. no falling behind check.

pkg/synchronizer/track.go
```diff
@@ -295,7 +306,7 @@ func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Dura
 		pts = t.lastPTS + t.toDuration(ts-t.lastTS)
 	}
 
-	if pts < t.lastPTS || !t.acceptable(pts-estimatedPTS) {
+	if pts < t.lastPTS || !t.acceptable(pts-estimatedPTS) || fallingBehind {
```
boks1971, line 309: please log `fallingBehind` so that we know the reason for entering this condition.

Also, in general, please consider logging all these things (state, what triggered condition etc) so that we have richer information.

pkg/synchronizer/track.go
```diff
@@ -264,15 +267,23 @@ func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Dura
...
+
 	if t.preJitterBufferReceiveTimeEnabled {
 		// if first packet a frame was too old and dropped,
-		// drop all packets of the frame irrespective of whether they are old or not
-		if ts == t.lastTSOldDropped || t.isPacketTooOld(pkt.ReceivedAt) {
+		// drop all packets of the frame irrespective of whether they are old or not, unless the track is falling behind
+		if (ts == t.lastTSOldDropped || t.isPacketTooOld(pkt.ReceivedAt)) && !fallingBehind {
 			t.lastTSOldDropped = ts
 			t.stats.numDroppedOld++
+			t.oldSequenceDuration += t.toDuration(ts - t.lastTS)
```
boks1971, line 286: not sure if this works. 

I think the problematic cases have RTP timestamp moving slower. The packet receive time is probably right. This will operate on packet receive time.

I think the diff in RTP timestamp needs to be measured over a long window and clock rate calculated using elapsed time. If that difference gets to be high, do the PTS adjustment.

milos-lk: you are right - I need to rework this.

milos-lk: the logic I have is getting really complicated (need to deal with mute/unmute and the edge case of continuous ts values after these) - I don't really like it. I think we can have a simpler solution for this - providing an optional callback that track syncrhonizer could use to get get pipeline running time (actually only a moment when pipeline switched to playing state). We can then evaluate timestamps we are crafting inside track syncrhonizer against that and make adjustments - that would be catch all kind of solution. wdyt?

boks1971: Yeah, that works. We can do that.

I think we can also do it without that. RTP time stamp should increment with elapsed time. So, it should jump across mute (there are some issues I have seen with React Native apps).

But, it is fine to use external clock to drive this. Maybe something like external clock reports a drift higher than some threshold continuously for certain amount of time. Wondering if we should do the start gate after every gap. Let us say a gap of 5 seconds re-enables the start gate (kind of like jitter buffer) so that we can actually keep sync better.

PR #766, opened 2025-10-13
Revive a track that constantly falls behind pipeline running time

milos-lk:
The second attempt to try to address issues with tracks whose calculated PTS continuously end up too far in the past comparing to the the media pipieline time.
Adding an option to set a callback to get a pipeline running time and max allowed latency comparing to it. If for whichever reason a track is falling behind the running time - latency continuoulsy for 10s - the workaround will pull that track back to the present moment.
Perfect sync can't be maintained after the rebasing but at least the track will be preserved.


pkg/synchronizer/track.go
```diff
@@ -117,6 +124,8 @@ func newTrackSynchronizer(s *Synchronizer, track TrackRemote) *TrackSynchronizer
 		enableStartGate:                   s.config.EnableStartGate,
 		nextPTSAdjustmentAt:               mono.Now(),
 		propagationDelayEstimator:         NewOWDEstimator(OWDEstimatorParamsDefault),
+		mediaRunningTime:                  s.config.MediaRunningTime,
+		maxMediaRunningTimeDelay:          s.config.MaxMediaRunningTimeDelay,
```
boks1971, line 131: log this config setting in `initialized track synchronizer` log please.

pkg/synchronizer/synchronizer.go
```diff
@@ -284,3 +301,31 @@ func (s *Synchronizer) GetEndedAt() int64 {
 
 	return s.endedAt
 }
+
+func (s *Synchronizer) getMediaPipelineDeadline() (time.Duration, bool) {
+	// take a snapshot of pipeline start under lock
```
boks1971: :nit: I usually advocate for not adding comments when the code is quite clear what it is doing.

pkg/synchronizer/synchronizer.go
```diff
@@ -135,6 +148,8 @@ type Synchronizer struct {
 	psByIdentity map[string]*participantSynchronizer
 	psBySSRC     map[uint32]*participantSynchronizer
 	ssrcByID     map[string]uint32
+
+	mediaPipelineStartTime time.Time
```
boks1971: can we avoid the name `pipeline` here? Can we use some name like external media clock? As this is Go SDK, I am hesitant to introduce more of egress concept like pipeline here. Basically, I want to think of this as some external clock impacting behaviour of synchroniser and that external clock source could be anything.

pkg/synchronizer/track.go
```diff
@@ -748,6 +758,49 @@ func (t *TrackSynchronizer) maybeAdjustStartTime(asr *augmentedSenderReport) int
 	return requestedAdjustment
 }
 
+func (t *TrackSynchronizer) normalizePTSToMediaPipelineTimeline(ptsIn time.Duration, ts uint32) (adjusted, ptsOut time.Duration) {
+	now := mono.Now()
```
boks1971: probably minor, but `mono.Now()` is already done at call site. Can pass it in.

pkg/synchronizer/track.go
```diff
@@ -748,6 +758,49 @@ func (t *TrackSynchronizer) maybeAdjustStartTime(asr *augmentedSenderReport) int
...
+	adjustedIn := ptsIn + t.currentPTSOffset
+	adjusted = adjustedIn
+	ptsOut = ptsIn
+
+	if t.sync == nil {
+		return adjusted, ptsOut
```
boks1971: can just do `return` if using named return to keep it consistent.

pkg/synchronizer/track.go
```diff
@@ -748,6 +758,49 @@ func (t *TrackSynchronizer) maybeAdjustStartTime(asr *augmentedSenderReport) int
...
+	}
+
+	deadline, ok := t.sync.getMediaPipelineDeadline()
+	if ok && adjustedIn < deadline {
+		if t.lastTimelyPacket.IsZero() {
+			t.lastTimelyPacket = now
+		}
+		if now.Sub(t.lastTimelyPacket) > cMaxTimelyPacketAge {
+			// track is constantlty behind, correct PTS to pull the track forward
```
boks1971: :nit: constantly spelling

pkg/synchronizer/track.go
```diff
@@ -748,6 +758,49 @@ func (t *TrackSynchronizer) maybeAdjustStartTime(asr *augmentedSenderReport) int
...
+	now := mono.Now()
+	adjustedIn := ptsIn + t.currentPTSOffset
+	adjusted = adjustedIn
+	ptsOut = ptsIn
+
+	if t.sync == nil {
+		return adjusted, ptsOut
+	}
+
+	deadline, ok := t.sync.getMediaPipelineDeadline()
```
boks1971: Is `getMediaPipelineDeadline()` expensive, i. e. getting it from gstreamer pipeline for example? should we do this once a second or something?

pkg/synchronizer/track.go
```diff
@@ -748,6 +758,49 @@ func (t *TrackSynchronizer) maybeAdjustStartTime(asr *augmentedSenderReport) int
...
+			)
+			ptsOut = newPTS
+			adjusted = ptsOut + t.currentPTSOffset
+			t.startRTP = newStartRTP
+			t.lastTimelyPacket = now
+		}
+	} else {
+		t.lastTimelyPacket = now
+	}
+	return adjusted, ptsOut
```
boks1971: just return as named return?

pkg/synchronizer/track.go
milos-lk: MediaRunningTime cb inside synchronizer is expensive (cgo) - that's why I added this method that establishes the mapping once and keep returning values based on that cached result. For live pipelines the running time will keep monotonically increasing - so mapping once is easy and probably sufficient for the use case. We can indeed keep periodically try to update it to make it a bit more general (e.g for the case where external pipeline doesn't use the same clock source).

pkg/synchronizer/synchronizer.go
milos-lk: updated

pkg/synchronizer/track.go
milos-lk: updated

pkg/synchronizer/track.go
milos-lk: 👍 done

pkg/synchronizer/track.go
boks1971: Got it. Thank you.

pkg/synchronizer/track.go
milos-lk: sharp eye -  I definitively need to add spelling linter.

pkg/synchronizer/track.go
milos-lk: updated

pkg/synchronizer/track.go
milos-lk: added

PR #771, opened 2025-10-18
Ensure adjusted PTS is always moving forward.

boks1971:
As adjusted is calculated after applying offset, it is possible that it could go backwards in some edge cases. Ensure it is always moving forward.

### DIFF
--- a/pkg/synchronizer/track.go
+++ b/pkg/synchronizer/track.go
@@ -335,6 +335,20 @@ func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Dura
 	}
 
 	adjusted := pts + t.currentPTSOffset
+	if adjusted < t.lastPTSAdjusted {
+		// always move it forward
+		t.logger.Infow(
+			"propelling PTS forward",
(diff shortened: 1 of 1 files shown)


pkg/synchronizer/track.go
```diff
@@ -335,6 +335,20 @@ func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Dura
 	}
 
 	adjusted := pts + t.currentPTSOffset
+	if adjusted < t.lastPTSAdjusted {
```
milos-lk, line 338: since we already ensure pts < last pts above - I was thinking that we should maybe update `shouldAdjustPTS` function to take into account the advancement since the last packet and if smaller than adjustment step + negative drift we return false and avoid adjustment?

boks1971: That sounds good

milos-lk: we can still leave this one just in case

milos-lk: This is what I had in mind:
https://github.com/livekit/server-sdk-go/pull/772

PR #775, opened 2025-10-21
Cloud Agents Client

yoonsio:
NOTE: we need to ensure livekit-cli major and minor version matches server-sdk-go release tag.

ref HA-304

### DIFF
--- a/pkg/cloudagents/logs.go
+++ b/pkg/cloudagents/logs.go
@@ -55,9 +55,8 @@ func (c *Client) StreamLogs(ctx context.Context, logType, agentID string, writer
 		var errorResponse APIError
 		if err := json.NewDecoder(resp.Body).Decode(&errorResponse); err != nil {
 			return fmt.Errorf("failed to parse error response: %w", err)
-		} else {
-			return fmt.Errorf("failed to get logs: %s", errorResponse.Message)
 		}
+		return fmt.Errorf("failed to get logs: %s - %v", errorResponse.Message, errorResponse.Meta)
 	}
 	scanner := bufio.NewScanner(resp.Body)
 	for {


README.md
```diff
@@ -373,6 +373,18 @@ func ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+This project shares the same **major** and **minor** version with [`livekit-cli`](https://github.com/livekit/livekit-cli).
+
+compatibility | server-sdk-go | livekit-cli
+--- | --- | ---
+✅ | 2.2.x | 2.2.x
+❌ | 2.1.x | 2.4.x
+❌ | 3.1.x | 2.1.x
+
+When bumping the major version, the `livekit-cli` verison must be updated accordingly.
```
rektdeckard: Can you add a note about ordering and consequences here? "A SDK/CLI major version release must precede update in cloud-agents service" or whatever needs to be known to implement this without making mistakes.

pkg/cloudagents/logs.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+type APIError struct {
+	Message string             `json:"msg"`
+	Meta    *map[string]string `json:"meta,omitempty"`
```
rektdeckard, line 30: map is already a reference type, so don't think the extra pointer is necessary

rektdeckard (approved):
Appears to be mostly porting the existing CLI code so I'm assuming no major change in functionality.

pkg/cloudagents/upload.go
```diff
@@ -0,0 +1,129 @@
...
+	"mime/multipart"
+	"net/http"
+
+	"github.com/livekit/protocol/livekit"
+)
+
+var (
+
+	//go:embed defaults/*
+	embedfs embed.FS
```
rektdeckard: We might want to expose this as part of the public API, I think we want to use some of these files for initial codegen if creating an agent with CLI that does not come from a template that already has these files included?

README.md
yoonsio: Actually, the minor and major version must always match from `livekit-cli` side. I removed the section from `server-sdk-go`.

rektdeckard: Right, they must always match -- but they also need to match (and be published) *before* cloud-agents implements the new version, correct?

pkg/cloudagents/logs.go
yoonsio: `omitempty` requires it to be a pointer otherwise it gets rendered

pkg/cloudagents/upload.go
yoonsio: Could you elaborate what we want to expose? The actual default ignore files?

rektdeckard: Yes, default .dockerignore files. I would also move the default Dockerfiles here too so they're co-located and easy to update together.

yoonsio: I am actually going to ignore it from here and expect dockerignore to exist in the file system. (generated by livekit-cli)

rektdeckard: I'd argue that if we're trying to centralize the cloud agents API here it should be batteries-included. Both CLI and cloud-api-server will now want to do things like initialize an agent project directory.

rektdeckard: But I can also see the argument for not embedding these files in everyone's copy of the SDK even when not using cloud agents

yoonsio: I think any project-specific file generation should be handled by the caller. Currently, only the CLI generates a Dockerfile. The server-sdk should only exclude files when necessary, based on the ignore rules.

pkg/cloudagents/logs.go
rektdeckard: Ah, i see. You want to differentiate between "abesnt" and "empty". But I don't see it being used anywhere? Anyway no issue.

rektdeckard: FWIW `omitempty` doesn't require a pointer here, it will skip both nil ptr and empty maps when marshaling, and when unmarshaling it will generate the appropriate type based on the field.

yoonsio: We should be showing metadata. Fixed.

pkg/cloudagents/build.go
```diff
@@ -0,0 +1,88 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	bkclient "github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/util/progress/progressui"
```
kuba--, line 29: ?

pkg/cloudagents/build.go
```diff
@@ -0,0 +1,88 @@
...
+	ch := make(chan *bkclient.SolveStatus)
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		display, err := progressui.NewDisplay(os.Stderr, "auto")
+		if err != nil {
+			return err
+		}
+		_, err = display.UpdateFrom(context.Background(), ch)
+		return err
+	})
```
kuba--, line 59: In what universe, do we need `ui` progress bar on server-sdk?
This stuff wastes resources:
```
if v := os.Getenv("TTY_DISPLAY_RATE"); v != "" {
		if r, err := strconv.ParseInt(v, 10, 64); err == nil {
			tickerTimeout = time.Duration(r) * time.Millisecond
			displayTimeout = time.Duration(r) * time.Millisecond
		}
	}
...
for {
		select {
		case <-ctx.Done():
			return nil, context.Cause(ctx)
		case <-ticker.C:
			d.disp.refresh()
		case ss, ok := <-ch:

```
and not giving any benefits here!

kuba--: Do we need any progress here, just take an interface and call it. If someone wants UI, then up to him.
On server side it should all work on callbacks.

pkg/cloudagents/client.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (c *Client) getAgentsURL(serverRegion string) string {
+	agentsURL := c.projectURL
+	if strings.HasPrefix(agentsURL, "ws") {
+		agentsURL = strings.Replace(agentsURL, "ws", "http", 1)
+	}
+	if os.Getenv("LK_AGENTS_URL") != "" {
+		agentsURL = os.Getenv("LK_AGENTS_URL")
+	} else if !strings.Contains(agentsURL, "localhost") && !strings.Contains(agentsURL, "127.0.0.1") {
```
kuba--, line 150: Shall we overwrite `agentsURL` here? I think env var. has a lowest priority?

pkg/cloudagents/client.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+func (c *Client) getAgentsURL(serverRegion string) string {
+	agentsURL := c.projectURL
+	if strings.HasPrefix(agentsURL, "ws") {
+		agentsURL = strings.Replace(agentsURL, "ws", "http", 1)
+	}
+	if os.Getenv("LK_AGENTS_URL") != "" {
+		agentsURL = os.Getenv("LK_AGENTS_URL")
```
kuba--, line 149: if we take it from env.var we don't validate like we do 👇🏻

pkg/cloudagents/logs.go
```diff
@@ -0,0 +1,82 @@
...
```
kuba--, line 30: IMO it's very error prone. I would stick to regular `map` (no pointer) - does not matter what `omitempty` requires. It's safer to remove `omitempty` and use regular map type, then other way around, IMO.

pkg/cloudagents/build.go
yoonsio: oh this is my bad. I thought I removed all progressbar. Will fix

pkg/cloudagents/build.go
yoonsio: Will remove!

pkg/cloudagents/client.go
yoonsio: good find. Will fix

pkg/cloudagents/logs.go
yoonsio: This is not inherently dangerous since json decoding never produces a case where Meta is not nil but underlying map is not. This is only way for a field to be nullable due to the limitation with omitempty. I will change this here because I am not relying on nullable behavior here.

pkg/cloudagents/client.go
yoonsio: The env always has highest priority but should still be validated.

pkg/cloudagents/build.go
yoonsio: We need to support streaming during the build.

The buildkit library uses an internal go channel to return build logs, but returning this raw channel directly to caller exposes internal concurrency details and makes channel lifecycle management difficult. It is ambiguous who is responsible for closing the channel, consuming messages, or handling potential blocking operations.

A better approach is to use `io.Writer` interface which allows the function to manage concurrency internally while providing a cleaner API.

The `progressui` package name can be misleading. It is essentially a concurrent writer with a ticker and optional formatting configuration. It can also be used to stream raw JSON data if needed. While we could implement our own ticker to read from the internal channel, using the `progressui` package is more maintainable solution.

yoonsio: I will fix the part where os.Stderr is hard-coded.

PR #782, opened 2025-10-22
cloudagents client interface to support build logs

yoonsio:
Address few additional comments for cloudagents client
* fix agentsURL override
* support build logs streaming

Comment on progressui from the [other PR](https://github.com/livekit/server-sdk-go/pull/775#discussion_r2453008297):

> We need to support streaming during the build.
> 
> The buildkit library uses an internal go channel to return build logs, but returning this raw channel directly to caller exposes internal concurrency details and makes channel lifecycle management difficult. It is ambiguous who is responsible for closing the channel, consuming messages, or handling potential blocking operations.
>
> A better approach is to use io.Writer interface which allows the function to manage concurrency internally while providing a cleaner API.
>
> The progressui package name can be misleading. It is essentially a concurrent writer with a ticker and optional formatting configuration. It can also be used to stream raw JSON data if needed. While we could implement our own ticker to read from the internal channel, using the progressui package is more maintainable solution.

### DIFF
--- a/pkg/cloudagents/client.go
+++ b/pkg/cloudagents/client.go
@@ -17,6 +17,7 @@ package cloudagents
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/fs"
 	"net/http"
 	"os"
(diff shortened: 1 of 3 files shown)


rektdeckard (approved):

pkg/cloudagents/client.go
```diff
@@ -76,6 +77,7 @@ func (c *Client) CreateAgent(
 	secrets []*lkproto.AgentSecret,
 	regions []string,
 	excludeFiles []string,
+	buildLogStreamWriter io.Writer,
 ) (*lkproto.CreateAgentResponse, error) {
```
kuba--, line 81: Btw. for server sdk, I wonder why build logs cannot go in response?

yoonsio: That is because it needs to be streamed while the build is running. The caller can just dump it to discard writer if not needed.

PR #787, opened 2025-10-27
fix(cloudagents): force venv/.venv to be omitted from tarballs

rektdeckard:


version.go
```diff
@@ -14,4 +14,4 @@
 
 package lksdk
 
-const Version = "2.12.1"
+const Version = "2.12.2"
```
yoonsio, line 17: I don't think this is urgent that we need to cut a new release.

rektdeckard: fair, but aren't we now sourcing only explicit versions in CLI? https://github.com/livekit/livekit-cli/blob/main/go.mod#L16

rektdeckard: if we want it there soon, we should bump. it's not breaking as far as cloud-agents is concerned anyway

PR #793, opened 2025-11-01
Add support for AVCC reader for H.264 publish.

boks1971:
Thanks to Sugun Segu@Oxa for the patch.

With AnnexB, when reading from socket, have to wait for H.264 start code. That adds a frame of delay as previous frame cannot be dispatched till the next frame's start code is available.

With AVCC, there is a header length prefix which allows reading a frame and dispatching it without waiting for next frame start. Streaming cameras support the RTP mode output format which allows this.

Ideally, we could get this mode of operation into pion/webrtc as well, but that would be longer route. As we want to make a CLI release with this, picking up the patch in the SDK.

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -41,18 +41,18 @@ const (
 
 // ---------------------------------
 
-type H264Format int
+type H26xFormat int
 
 const (
-	H264FormatAnnexB H264Format = iota
-	H264FormatAVCC
+	H26xFormatAnnexB H26xFormat = iota
+	H26xFormatAVCC
 )
 
-func (f H264Format) String() string {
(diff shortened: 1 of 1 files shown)


localtrack.go
```diff
@@ -312,32 +312,38 @@ func (s *LocalTrack) WriteRTP(p *rtp.Packet, opts *SampleWriteOptions) error {
 	s.lock.RLock()
 	transceiver := s.transceiver
 	ssrcAcked := s.ssrcAcked
+	audioLevelID := s.audioLevelID
```
boks1971, line 315: CI reported data race. Fixing that as well.

milos-lk (approved):

PR #797, opened 2025-11-03
Use build log stream auto format by default

yoonsio:
Using 'auto' format allows the build logs to roll up in TTY

### DIFF
--- a/pkg/cloudagents/build.go
+++ b/pkg/cloudagents/build.go
@@ -47,7 +47,7 @@ func (c *Client) build(ctx context.Context, id string, writer io.Writer) error {
 		return fmt.Errorf("failed to build agent: %s", resp.Status)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloudagents/build.go
```diff
@@ -47,7 +47,7 @@ func (c *Client) build(ctx context.Context, id string, writer io.Writer) error {
 		return fmt.Errorf("failed to build agent: %s", resp.Status)
 	}
 
-	displayMode := progressui.PlainMode
```
rektdeckard, line 50: I think this will break the new streaming logs API we need for web. `AutoMode` will do `tty` if the writer supports it, otherwise will fall back to `plain`. We want `rawjson` for that API.

rektdeckard: Sorry, nvm I see the line below XX

PR #804, opened 2025-11-12
Introduced OnParticipantDisconnectedWithReason

beelis:
Introduced `OnParticipantDisconnectedWithReason` since the reason is not set in the `RemoteParticipant` received in the `OnParticipantDisconnected` callback.

### DIFF
--- a/room.go
+++ b/room.go
@@ -809,7 +809,7 @@ func (r *Room) OnRestarting() {
 	r.callback.OnReconnecting()
 
 	for _, rp := range r.GetRemoteParticipants() {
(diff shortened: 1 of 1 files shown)


anunaym14:
@davidzhao @beelis something like: https://github.com/livekit/server-sdk-go/pull/805 seems to work as well, do we need a separate callback for this?

beelis:
> @davidzhao @beelis something like: #805 seems to work as well, do we need a separate callback for this?

@anunaym14 totally fine by me, as long as i'm somehow able to get the reason i'm happy 😊 Thank you!

PR #812, opened 2025-12-02
chore(release): v2.13.0

anunaym14:


biglittlebigben (approved):
Could you run a `go mod tidy` to remove the unused entries from go.sum?

PR #813, opened 2025-12-04
Do not overwrite the exsting Twirp headers on the context in the auth middleware

biglittlebigben:


boks1971 (approved):

auth.go
```diff
@@ -61,5 +62,15 @@ func (b authBase) withAuth(ctx context.Context, opt authOption, options ...authO
...
 	}
 
-	return twirp.WithHTTPRequestHeaders(ctx, signalling.NewHTTPHeaderWithToken(token))
+	h := signalling.NewHTTPHeaderWithToken(token)
+	ctxH, _ := twirp.HTTPRequestHeaders(ctx)
+
+	// merge new header with the ones already present in the context
+	for k, vv := range ctxH {
+		for _, v := range vv {
+			h.Add(k, v)
```
dennwc: This might potentially overwrite auth headers. Do we want to do it the other way around?

dennwc (approved):

PR #817, opened 2025-12-18
Pass Interceptors to Subscriber PeerConnection for Stats Collection

terran-vigil-simplisafe:
The createSubscriberPCLocked does not pass Interceptors from e.connParams to the subscriber PeerConnection. Note that createPublisherPCLocked does do this.
This prevents interceptors (in our case, Pion WebRTC stats interceptor) from being applied to the subscriber, causing inbound-rtp statistics to be missing when GetStats() is called on  the subscriber PeerConnection.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -73,12 +73,12 @@ type PCTransportParams struct {
 	Configuration webrtc.Configuration
 	Codecs        []webrtc.RTPCodecParameters
 
-	RetransmitBufferSize uint16
-	Pacer                pacer.Factory
-	Interceptors         []interceptor.Factory
+	RetransmitBufferSize       uint16
+	Pacer                      pacer.Factory
+	Interceptors               []interceptor.Factory
 	IncludeDefaultInterceptors bool
-	OnRTTUpdate          func(rtt uint32)
-	IsSender             bool
+	OnRTTUpdate                func(rtt uint32)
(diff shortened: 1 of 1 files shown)


terran-vigil-simplisafe:
> LGTM, but I just want to double-check if this was done intentionally

Yes, it was intentional. It was the only way I could see to capture the metrics I needed.

anunaym14:
> Yes, it was intentional. It was the only way I could see to capture the metrics I needed.

Sorry, I meant if leaving it out of subscriber peer connection in the first place was intentional or not

transport.go
```diff
@@ -76,6 +76,7 @@ type PCTransportParams struct {
 	RetransmitBufferSize uint16
 	Pacer                pacer.Factory
 	Interceptors         []interceptor.Factory
+	IncludeDefaultInterceptors bool
```
boks1971, line 79: looks good overall,

this formatting looks off, can you please check that?

terran-vigil-simplisafe: Done. Thanks

terran-vigil-simplisafe: Anything else need to be done?

boks1971: I think this is good to go.

terran-vigil-simplisafe:
Thanks!

PR #825, opened 2026-01-13
Add User Timestamp trailer support

chenosaurus:


readersampleprovider.go
```diff
@@ -292,6 +308,20 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
 			nalUnitData = nal.Data
 		}
 
+		if nalUnitType == h264reader.NalUnitTypeSEI {
+			if p.appendUserTimestamp {
+				if ts, ok := parseH264SEIUserTimestamp(nalUnitData); ok {
+					p.pendingUserTimestampUs = ts
```
boks1971, line 341: looks like these two variables does not need to be in the struct, can just be local?

user_timestamp.go
```diff
@@ -0,0 +1,57 @@
...
+//   - 4-byte ASCII magic "LKTS"
+func appendUserTimestampTrailer(data []byte, userTimestampUs int64) []byte {
+	outLen := len(data) + userTimestampTrailerSize
+	out := make([]byte, outLen)
+	copy(out, data)
+
+	// Write timestamp (big-endian) just before the magic bytes.
+	tsOffset := len(data)
+	var tsBuf [8]byte
+	binary.BigEndian.PutUint64(tsBuf[:], uint64(userTimestampUs))
```
boks1971: directly write to `out` at `len(data)`?

boks1971 (approved):

readersampleprovider.go
chenosaurus: `pendingUserTimestampUs` and `hasPendingUserTimestamp` does need to be on the struct.  When we parse a SEI NAL, we extract the timestamp and store it to be used in the next call of `NextSample` that returns a video frame and gets attached to that frame.

boks1971: sorry my bad, missed the early return on SEI NALu.

PR #826, opened 2026-01-13
wav support

biglittlebigben:


readersampleprovider.go
```diff
@@ -166,6 +172,15 @@ func NewLocalFileTrack(file string, options ...ReaderSampleProviderOption) (*Loc
 		mime = webrtc.MimeTypeH265
 	case ".ogg":
 		mime = webrtc.MimeTypeOpus
+	case ".wav":
+		// Parse WAV header to determine format (PCMU or PCMA)
+		var err error
+		mime, err = detectWavFormat(fp)
+		if err != nil {
+			_ = fp.Close()
+			return nil, fmt.Errorf("failed to detect WAV format: %w", err)
```
biglittlebigben: We should create a new error constant or reuse ErrCannotDetermineMime

readersampleprovider.go
```diff
@@ -241,6 +258,19 @@ func (p *ReaderSampleProvider) OnBind() error {
 		}
 	case webrtc.MimeTypeOpus:
 		p.oggReader, _, err = oggreader.NewOggReader(p.reader)
+	case webrtc.MimeTypePCMU, webrtc.MimeTypePCMA:
+		// wavReader requires io.ReadSeeker, check if reader supports it
+		readSeeker, ok := p.reader.(io.ReadSeeker)
```
biglittlebigben: Can we avoid the type assertion here? Do we really need the seek? I yes, can we make reader a "ReadSeekerCloser"?

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+	var channels uint16
+	var bitsPerSample uint16
+	var dataOffset int64
+	var dataSize int64
+
+	pos := int64(12)
+	for {
+		// Read chunk header
+		var chunkHeader [8]byte
+		if _, err := r.Seek(pos, io.SeekStart); err != nil {
```
biglittlebigben: I don't think we need to seek here?

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+			if chunkSize < 16 {
+				return nil, "", fmt.Errorf("fmt chunk too small: %d bytes (minimum 16)", chunkSize)
+			}
+
+			// Cap chunk size to prevent malicious/malformed files from causing huge allocations
+			// We only parse the first 16 bytes, so 64 bytes is more than enough
+			const maxFmtChunkSize = 64
+			readSize := chunkSize
+			if readSize > maxFmtChunkSize {
+				readSize = maxFmtChunkSize
```
biglittlebigben: We need to return an error, not keep reading a malformed file

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+			}
+
+			// Safely convert int64 to int for make(), guarding against overflow
+			if readSize > int64(^uint(0)>>1) {
```
biglittlebigben: We just checked this as < 64...

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+				return nil, "", fmt.Errorf("fmt chunk size too large: %d bytes (max %d)", chunkSize, maxFmtChunkSize)
+			}
+
+			// Read format chunk (read up to maxFmtChunkSize bytes, but we only parse first 16)
+			fmtChunk := make([]byte, int(readSize))
+			if _, err := r.Seek(chunkDataStart, io.SeekStart); err != nil {
```
biglittlebigben: Do we need the Seek?

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+			}
+			if _, err := io.ReadFull(r, fmtChunk); err != nil {
+				return nil, "", fmt.Errorf("failed to read fmt chunk: %w", err)
+			}
+
+			// Parse required fields (first 16 bytes)
+			formatCode = binary.LittleEndian.Uint16(fmtChunk[0:2])
+			channels = binary.LittleEndian.Uint16(fmtChunk[2:4])
+			sampleRate = binary.LittleEndian.Uint32(fmtChunk[4:8])
+			bitsPerSample = binary.LittleEndian.Uint16(fmtChunk[14:16])
```
biglittlebigben: I trust this is the right header format... I can't tell without looking up the spec...

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+			if sampleRate != 8000 {
+				return nil, "", fmt.Errorf("expected 8000 Hz sample rate for G.711, got %d", sampleRate)
+			}
+		} else if chunkID == "data" {
+			dataOffset = chunkDataStart
+			dataSize = chunkSize
+		}
+
+		// Move to next chunk (chunk size is rounded up to even boundary)
+		pos = chunkDataStart + chunkSize
```
biglittlebigben: Should we just error out if the chunk size is odd? If odd sizes are valid, we should just read 1 extra byte instead of calling Seek...

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+	if formatCode == wavFormatPCMU {
+		mime = webrtc.MimeTypePCMU
+	} else if formatCode == wavFormatPCMA {
+		mime = webrtc.MimeTypePCMA
+	} else {
+		return nil, "", fmt.Errorf("unsupported WAV format code: 0x%04x (expected PCMU 0x0007 or PCMA 0x0006)", formatCode)
+	}
+
+	// Seek to start of data
+	if _, err := r.Seek(dataOffset, io.SeekStart); err != nil {
```
biglittlebigben: Loop above is reading the whole wav file (for no reason), and then seeking back to data start. We should either stop reading after the header or copy all data in a buffer.

readersampleprovider.go
```diff
@@ -379,6 +424,189 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+		samplesPerFrame: defaultG711SamplesPerFrame,
+		bytesRead:       0,
+	}, mime, nil
+}
+
+// readFrame reads one frame of G.711 audio (160 samples = 20ms at 8kHz)
+// It respects dataSize and stops reading at the end of the audio data chunk
+func (w *wavReader) readFrame() ([]byte, error) {
+	// Check if we've read all available data
+	remaining := w.dataSize - w.bytesRead
```
biglittlebigben, line 556: We probably don't need this dataSize test and can rely on the reader returning EOF

readersampleprovider.go
```diff
@@ -398,3 +626,28 @@ func nextNALH264LengthPrefixed(r io.Reader) (h264reader.NalUnitType, []byte, err
...
+	currentPos, err := r.Seek(0, io.SeekCurrent)
+	if err != nil {
+		return "", fmt.Errorf("failed to get current position: %w", err)
+	}
+
+	// Parse WAV header to get mime type
+	_, mime, err := newWavReader(r)
+	if err != nil {
+		// Try to restore position even on error
+		_, _ = r.Seek(currentPos, io.SeekStart)
```
biglittlebigben: Do we need to restore the position?

readersampleprovider.go
biglittlebigben: 1 byte padding is actually part of the spec. The Seek is still a bit odd though.

readersampleprovider.go
milos-lk: no need for seeking - reusing the same wav reader between checking mime type and reading eliminates the need for seeking

readersampleprovider.go
milos-lk: not needed

readersampleprovider.go
milos-lk: correct

readersampleprovider.go
milos-lk: yeah - I simplified that here: https://github.com/livekit/server-sdk-go/pull/827

readersampleprovider.go
milos-lk: no need

readersampleprovider.go
milos-lk: in practice probably we don't need to do this - theoretically data doesn't need to be the last chunk - we can leave it

milos-lk (approved):

PR #827, opened 2026-01-14
wav reader unvibing

milos-lk:
simplifying wav reader implementation

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -133,6 +133,12 @@ func ReaderTrackWithH26xStreamingFormat(h26xStreamingFormat H26xStreamingFormat)
 	}
 }
 
+func readerTrackWithWavReader(wr *wavReader) func(provider *ReaderSampleProvider) {
+	return func(provider *ReaderSampleProvider) {
+		provider.wavReader = wr
+	}
+}
+
 // NewLocalFileTrack creates an *os.File reader for NewLocalReaderTrack
 func NewLocalFileTrack(file string, options ...ReaderSampleProviderOption) (*LocalTrack, error) {
 	// File health check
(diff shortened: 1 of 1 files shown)


readersampleprovider.go
```diff
@@ -440,139 +436,117 @@ const (
...
 	if _, err := io.ReadFull(r, riffHeader[:]); err != nil {
 		return nil, "", fmt.Errorf("failed to read RIFF header: %w", err)
 	}
 
-	// Check RIFF signature
 	if string(riffHeader[0:4]) != "RIFF" {
 		return nil, "", fmt.Errorf("not a RIFF file")
 	}
 
-	// Check WAVE signature
```
biglittlebigben, line 455: Should we read and validate the chunk size at the RIFF layer?

readersampleprovider.go
biglittlebigben, line 564: This will return an invalid frame. Should we error out instead?

readersampleprovider.go
```diff
@@ -440,139 +436,117 @@ const (
...
+				mime = webrtc.MimeTypePCMU
+			case wavFormatPCMA:
+				mime = webrtc.MimeTypePCMA
+			default:
+				return nil, "", fmt.Errorf("unsupported WAV format code: 0x%04x (expected PCMU 0x0007 or PCMA 0x0006)", formatCode)
+			}
+
+			return &wavReader{
+				reader:          r,
+				dataSize:        chunkSize,
```
biglittlebigben, line 524: nit: I think we may be able to replace dataSize and bytesRead with bytesRemaining?

biglittlebigben (commented):
LGTM. Couple of nits that are frankly not very important considering this is used as test code...

readersampleprovider.go
milos-lk: remaining is calculated based on data size - we can store it instead of data size but doesn't make any diff as far as can see. will leave it like this. 
Another PR needs to be added on top of this to enable PCMA/U tests

PR #828, opened 2026-01-16
Adding support to restrict codecs for publishing

milos-lk:
Adding WithCodec option to restrict advertised codecs to a specific codec type when publishing tracks 
- Refactor codec preference handling for consistency across all track publishing methods 
- Ensure codec restrictions are preserved when tracks are republished (e.g., after reconnection)                  

Motivated by adding PCMA/U codec support inside egress, for some of newly added tests although a local wav file track is created containing only PCMA (or PCMU) data - list of advertised codecs included opus as well - and it was sometimes selected by SFU only to follow with `dropping packet - payload mismatch` as the publisher was trying to send PCMA instead of opus. Maybe there is a better way to have only one codec selected (for file tracks - the native codec present there - if so please advise. I opted with WithCodec as it was easier for me to conceptually comprehend it.

Internal improvements: 
- Extract filterCodecPreferences() and reorderCodecPreferences() helper functions to replace duplicated codec manipulation logic
- Apply consistent codec handling across PublishTrack, PublishSimulcastTrack, publishAdditionalCodecForTrack, and              
  republishTracks 
- Store restrictCodec and codecPreference on LocalTrackPublication to preserve settings through republish cycles

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -142,8 +142,6 @@ func readerTrackWithWavReader(wr *wavReader) func(provider *ReaderSampleProvider
 // NewLocalFileTrack creates an *os.File reader for NewLocalReaderTrack
 func NewLocalFileTrack(file string, options ...ReaderSampleProviderOption) (*LocalTrack, error) {
 	// File health check
-	logger.Infow("NewLocalFileTrack", "file", file)
-	defer logger.Infow("NewLocalFileTrack done", "file", file)
 	var err error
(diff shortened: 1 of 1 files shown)


biglittlebigben (commented):
How does this new behavior differ from what we get by exporting `withCodecs` on the room? https://github.com/livekit/server-sdk-go/blob/d2e7028eedc1ac66bcad1d7d212a8b91ed1fa688/room.go#L180

Are we reimplementing logic already present in pion?

PR #837, opened 2026-01-27
after parsing H264 SEI NAL, discard to avoid breaking certain decoders

chenosaurus:
- when SEI parsing is enabled, we should always discard the SEI NAL as it still breaks some decoders.  

## Summary by CodeRabbit

* **Bug Fixes**
  * SEI metadata is no longer emitted as standalone samples; SEI content is always cleared and used only to source timestamps to prevent invalid/empty samples and potential decoder issues.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -379,7 +379,6 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
 				p.pendingUserTimestampUs = 0
 			}
 			sample.Data = appendUserTimestampTrailer(sample.Data, ts)
-
 		}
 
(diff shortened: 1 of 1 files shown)


readersampleprovider.go
```diff
@@ -378,6 +379,7 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
 				p.pendingUserTimestampUs = 0
 			}
 			sample.Data = appendUserTimestampTrailer(sample.Data, ts)
+
```
boks1971: :nit: delete empty line

PR #841, opened 2026-02-04
Fix for deadlock on inverse locking order

milos-lk:
Fix potential deadlock between Synchronizer.End() and TrackSynchronizer.GetPTS() caused by inverse lock ordering.
getPTSWithoutRebase() and getPTSWithRebase() now release the TrackSynchronizer lock before calling into Synchronizer methods

  Problem

  A deadlock could occur when two goroutines acquired locks in inverse order:

```
  Thread A (End()): Synchronizer.Lock() x
  Thread B (GetPTS()):
  ────────────────────────────────────────
  Thread A (End()):
  Thread B (GetPTS()): TrackSynchronizer.Lock() x
  ────────────────────────────────────────
  Thread A (End()): → p.getMaxOffset() → t.Lock() BLOCKED
  Thread B (GetPTS()):
  ────────────────────────────────────────
  Thread A (End()):
  Thread B (GetPTS()): → getExternalMediaDeadline() → s.RLock() BLOCKED
  Lock order in End(): Synchronizer → participantSynchronizer → TrackSynchronizer
```

  Lock order in GetPTS(): TrackSynchronizer → Synchronizer (inverse)

Going forward we need to avoid invocking any actions on syncrhonizer from the track synchronizer while holding its lock. Following that rule will ensure no lock inversion deadlock could happen.

## Release Notes

* **Bug Fixes**
  * Optimized media synchronization by improving how external timing deadlines are processed and integrated into presentation timestamp calculations.
  * Enhanced deadline-aware timestamp normalization to better account for media pipeline constraints and timing requirements throughout the synchronization process.
  * Strengthened overall synchronization reliability to ensure consistent, predictable playback and timing accuracy across multiple simultaneous media streams.

### DIFF
--- a/pkg/synchronizer/track.go
+++ b/pkg/synchronizer/track.go
@@ -250,6 +250,16 @@ func (t *TrackSynchronizer) GetPTS(pkt jitter.ExtPacket) (time.Duration, error) 
 
 func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Duration, error) {
 	t.Lock()
+	sync := t.sync
+	t.Unlock()
+
+	var deadline time.Duration
+	var hasDeadline bool
+	if sync != nil {
+		deadline, hasDeadline = sync.getExternalMediaDeadline()
+	}
+
+	t.Lock()
 	defer t.Unlock()
 
 	now := time.Now()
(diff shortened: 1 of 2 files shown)


pkg/synchronizer/track.go
```diff
@@ -249,6 +249,16 @@ func (t *TrackSynchronizer) GetPTS(pkt jitter.ExtPacket) (time.Duration, error)
 }
 
 func (t *TrackSynchronizer) getPTSWithoutRebase(pkt jitter.ExtPacket) (time.Duration, error) {
+	t.Lock()
```
boks1971, line 252: make a helper function and use it in both places?

milos-lk: good point - there is a bunch or repetition in these  methods we could tidy - let me check what we can simplify

boks1971: sorry, that is my bad, I wanted to keep the old code as unmodified as possible when we were adding OWD stuff just to make sure that the existing code does not get affected. I was also thinking that we will get rid of one of these paths soon. Consolidating sounds good.

milos-lk: hehe - no worries - I felt in the same duplication trap :)

PR #844, opened 2026-02-06
Handle H265 encoders that produce multi-slice IDR 

chenosaurus:
Fix issue where a h265 encoder outputs multi-slice IDR, causing downstream decoders to fail.  
- add detection for multi-slide IDR via `first_slice_segment_in_pic_flag` and collect the slices into a single AU before emitting the frame.

### DIFF
--- a/readersampleprovider.go
+++ b/readersampleprovider.go
@@ -15,6 +15,7 @@
 package lksdk
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"fmt"
@@ -749,44 +750,44 @@ func h265FirstSliceInPic(nalData []byte) (bool, bool) {
 // owned Annex-B buffer once we need to join multiple NAL units together.
 type h265AccessUnitBuilder struct {
 	rawFirstNAL []byte
-	data        []byte
+	data        bytes.Buffer
 }
 
 func (b *h265AccessUnitBuilder) Append(nalData []byte) {
 	// Preserve the single-NAL fast path without copying until another NAL forces
 	// us to materialize Annex-B framing.
-	if b.rawFirstNAL == nil && len(b.data) == 0 {
+	if b.rawFirstNAL == nil && b.data.Len() == 0 {
 		b.rawFirstNAL = nalData
 		return
 	}
 
 	b.materializeAnnexB(len(nalData) + len(annexBStartCode))
-	b.data = append(b.data, annexBStartCode[:]...)
-	b.data = append(b.data, nalData...)
+	_, _ = b.data.Write(annexBStartCode[:])
+	_, _ = b.data.Write(nalData)
 }
 
 func (b *h265AccessUnitBuilder) AppendAnnexB(nalData []byte) {
 	b.materializeAnnexB(len(nalData) + len(annexBStartCode))
-	b.data = append(b.data, annexBStartCode[:]...)
-	b.data = append(b.data, nalData...)
+	_, _ = b.data.Write(annexBStartCode[:])
+	_, _ = b.data.Write(nalData)
 }
 
 func (b *h265AccessUnitBuilder) Bytes() []byte {
 	if b.rawFirstNAL != nil {
 		return b.rawFirstNAL
 	}
-	return b.data
+	return b.data.Bytes()
 }
 
 func (b *h265AccessUnitBuilder) Len() int {
 	if b.rawFirstNAL != nil {
 		return len(b.rawFirstNAL)
 	}
-	return len(b.data)
+	return b.data.Len()
 }
 
 func (b *h265AccessUnitBuilder) materializeAnnexB(extra int) {
-	if b.rawFirstNAL == nil && b.data != nil {
+	if b.rawFirstNAL == nil && b.data.Len() != 0 {
 		if extra > 0 {
 			b.grow(extra)
 		}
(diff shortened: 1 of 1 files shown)


readersampleprovider.go
```diff
@@ -691,3 +730,36 @@ func detectWavFormat(r io.Reader) (*wavReader, string, error) {
...
+	}
+	return (nalData[2] & 0x80) != 0, true
+}
+
+func appendH265NAL(dst []byte, nalData []byte, sampleIsAnnexB *bool) []byte {
+	// Ensure AnnexB start codes and append in-place for AU aggregation.
+	if *sampleIsAnnexB || len(dst) > 0 {
+		if !*sampleIsAnnexB && len(dst) > 0 {
+			dst = append([]byte{0, 0, 0, 1}, dst...)
+			*sampleIsAnnexB = true
```
boks1971: should there be a return here? Not able to fully understand the sequence needed here, but letting this fall through seems like it adds another start code and also sets `sampleIsAnnexB` to `true` again. Seems redundant.

readersampleprovider.go
```diff
@@ -385,77 +387,114 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+					sample.Duration = 0
+					return sample, nil
+				}
+				continue
 			}
 
-			// aggregate vps,sps,pps into a single AP packet (chrome requires this)
+			// Aggregate VPS/SPS/PPS before the next access unit.
 			if nal.NalUnitType == 32 || nal.NalUnitType == 33 || nal.NalUnitType == 34 {
+				sampleIsAnnexB = true
```
boks1971: `sampleAnnexB` is also getting set here. Would be good to consolidate `AnnexB` determination in one place if possible?

readersampleprovider.go
```diff
@@ -385,77 +387,114 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+					break
+				}
 				return sample, err
 			}
 
+			if haveVCL {
+				// Once we've started a picture, detect the next access-unit boundary.
+				if nal.NalUnitType < 32 {
+					// VCL: split when first_slice_segment_in_pic_flag starts a new picture.
+					if isFirstSlice, ok := h265FirstSliceInPic(nal.Data); ok && isFirstSlice {
```
boks1971: can this be simplified a bit? looks like only the combination of `firstInSlice: true && ok: false` is the only one continuing. Can this be written as one check?
```
if isFirstSlice, ok := h265FirstSliceInPic(nal.Data); !ok || isFirstSlice
```
Did I get that logic right?

readersampleprovider.go
```diff
@@ -385,77 +387,114 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
+						// If we can't parse the flag, err on the side of splitting.
+						p.h265PendingNAL = nal
+						break
+					}
+				} else {
+					// Non-VCL after VCL belongs to the next access unit.
+					switch nal.NalUnitType {
+					case 40: // suffix SEI, ignore
+						continue
+					case 39, 32, 33, 34:
```
boks1971: Is this case different from `default` below? If not, can this just be made a comment and made let the `default` be the handler?

boks1971 (approved):
A few comments @chenosaurus . LGTM, but I am not familiar with all the nuances of the bitstream. Getting some more 👀 on this would be great.

readersampleprovider.go
chenosaurus: ur right, simplified

readersampleprovider.go
chenosaurus: cleaned up

readersampleprovider.go
```diff
@@ -385,77 +387,103 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
-				sample.Data = append(sample.Data, nal.Data...)
-			} else {
-				sample.Data = nal.Data
-			}
+			// Aggregate VPS/SPS/PPS and VCL NALs into the current access unit.
+			sample.Data = appendH265NAL(sample.Data, nal.Data, &sampleIsAnnexB)
 
-			if !isFrame {
-				isFrame = nal.NalUnitType < 32
+			if nal.NalUnitType == 32 || nal.NalUnitType == 33 || nal.NalUnitType == 34 {
```
biglittlebigben, line 458: Can we comment what these NAL IDs represent?

readersampleprovider.go
```diff
@@ -691,3 +719,39 @@ func detectWavFormat(r io.Reader) (*wavReader, string, error) {
...
+}
+
+func appendH265NAL(dst []byte, nalData []byte, sampleIsAnnexB *bool) []byte {
+	// Ensure AnnexB start codes and append in-place for AU aggregation.
+	if *sampleIsAnnexB || len(dst) > 0 {
+		if !*sampleIsAnnexB {
+			// First NAL was appended raw; retroactively add its start code.
+			dst = append([]byte{0, 0, 0, 1}, dst...)
+		}
+		dst = append(dst, []byte{0, 0, 0, 1}...)
```
biglittlebigben: This is going to be very inefficient. Can we use a bytes.Buffer instead?

PR #858, opened 2026-03-11
fix: additional act support

elizabeth-muirhead-simplisafe:
Related to https://github.com/livekit/livekit/pull/4243, and https://github.com/livekit/livekit/issues/4325. This pr does the following:
- Register ACT in MediaEngine for audio during negotiation.
- Capture negotiated ACT extension ID in LocalTrack.Bind().
- Remap ACT extension ID in LocalTrack.WriteRTP() for relay cases where source/destination PeerConnections use different ext IDs.

### DIFF
--- a/localtrack.go
+++ b/localtrack.go
@@ -323,9 +323,8 @@ func (s *LocalTrack) OnUnbind(f func()) {
 	s.lock.Unlock()
 }
 
-// SetIncomingAbsCaptureTimeExtensionID sets the incoming Absolute Capture Time extension ID.
-// This is used when relaying RTP packets between PeerConnections where incoming and
-// outgoing negotiated extension IDs may differ.
+// This is used to translate abs-capture-time extension before writing to a peer connection when source provided
+// extension ID does not match negotiated extension ID on the peer connection
 func (s *LocalTrack) SetIncomingAbsCaptureTimeExtensionID(id uint8) {
 	s.lock.Lock()
 	s.incomingAbsCaptureTimeID = id


localtrack.go
```diff
@@ -337,6 +344,22 @@ func (s *LocalTrack) WriteRTP(p *rtp.Packet, opts *SampleWriteOptions) error {
 		}
 	}
 
+	if s.Kind() == webrtc.RTPCodecTypeAudio && s.absCaptureTimeID != 0 && p.Header.Extension {
```
boks1971: why limit to only audio? Can be for video also.

localtrack.go
```diff
@@ -337,6 +344,22 @@ func (s *LocalTrack) WriteRTP(p *rtp.Packet, opts *SampleWriteOptions) error {
...
+		if data := p.Header.GetExtension(s.absCaptureTimeID); len(data) >= absCaptureTimeSize {
```
boks1971: empty `if`, what is the purpose of this?

localtrack.go
```diff
@@ -337,6 +344,22 @@ func (s *LocalTrack) WriteRTP(p *rtp.Packet, opts *SampleWriteOptions) error {
...
+		} else {
+			for _, id := range p.Header.GetExtensionIDs() {
+				if id == s.absCaptureTimeID {
+					continue
+				}
+				if data := p.Header.GetExtension(id); len(data) == 8 || len(data) == 16 {
```
boks1971: I don't think this would work. There could be other extensions with that length.

The right approach would be to record incoming abs capture time ID. If there is a need for an API for that, we should consider that and do translation of reading incoming and translating to outgoing.

transport.go
```diff
@@ -161,6 +161,10 @@ func NewPCTransport(params PCTransportParams) (*PCTransport, error) {
 	if err := m.RegisterHeaderExtension(sdesRtpStreamIdExtension, webrtc.RTPCodecTypeVideo); err != nil {
 		return nil, err
 	}
+	absCaptureTimeExtension := webrtc.RTPHeaderExtensionCapability{URI: "http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time"}
+	if err := m.RegisterHeaderExtension(absCaptureTimeExtension, webrtc.RTPCodecTypeAudio); err != nil {
+		return nil, err
```
boks1971, line 166: 1. This can be registered for video also.
2. Can use the constant defined in `localtrack.go` for the URI.

localtrack.go
elizabeth-muirhead-simplisafe: I updated this to support both media types

localtrack.go
elizabeth-muirhead-simplisafe: I removed the statement and simplified the flow so remapping only runs when explicit incoming/outgoing ACT IDs are set and different

transport.go
elizabeth-muirhead-simplisafe: I created a constant, and added a block for video.

localtrack.go
elizabeth-muirhead-simplisafe: I replaced the length-based guess with explicit ACT extension ID mapping from the incoming track to the negotiated outgoing ID

localtrack.go
```diff
@@ -316,12 +323,23 @@ func (s *LocalTrack) OnUnbind(f func()) {
 	s.lock.Unlock()
 }
 
+// SetIncomingAbsCaptureTimeExtensionID sets the incoming Absolute Capture Time extension ID.
+// This is used when relaying RTP packets between PeerConnections where incoming and
```
boks1971: :nit: this is not necessarily relaying between PeerConnections. Input could be from any source. Can just change the comment to something like `This is used to translate abs-capture-time extension before writing to a peer connection when source provided extension ID does not match negotiated extension ID on the peer connection`

PR #860, opened 2026-03-15
Rename log field pID to participantID

paulwe:
## Summary
- Rename all occurrences of the `"pID"` log field to `"participantID"` for consistency across repos

## Test plan
- [ ] Verify logs emit `participantID` instead of `pID`


localparticipant.go
```diff
@@ -1063,6 +1063,7 @@ func (p *LocalParticipant) StreamBytes(options StreamBytesOptions) *ByteStreamWr
 	writer := newByteStreamWriter(info, header, p.engine, options.DestinationIdentities, options.OnProgress)
 
 	p.engine.OnClose(func() {
+		// XXX
```
boks1971: claude adding random stuff

paulwe: no, that's me...

PR #868, opened 2026-03-25
Fix data race contributing to wrong value of ended_at

milos-lk:
Fix `endedAt == startedAt` (zero duration) when all tracks are removed before `Synchronizer.End()` is called.

This is a race in the egress shutdown path: track workers call `RemoveTrack` (which deletes from `p.tracks`) before the pipeline calls `End()`. When `End()` iterates participants to find `maxPTS`, the tracks map is already empty, so `maxPTS = 0` and `endedAt = startedAt`.

The bug was latent before the duration inflation fix but masked by `time.Now()` fallback.

### DIFF
--- a/pkg/synchronizer/track.go
+++ b/pkg/synchronizer/track.go
@@ -229,6 +229,12 @@ func (t *TrackSynchronizer) initialize(extPkt jitter.ExtPacket) {
 	)
 }
 
+func (t *TrackSynchronizer) LastPTSAdjusted() time.Duration {
+	t.Lock()
+	defer t.Unlock()
+	return t.lastPTSAdjusted
+}
+
 func (t *TrackSynchronizer) Close() {
 	t.Lock()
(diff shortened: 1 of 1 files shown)


pkg/synchronizer/participant.go
```diff
@@ -43,11 +44,26 @@ func (p *participantSynchronizer) onSenderReport(pkt *rtcp.SenderReport) {
 	}
 }
 
+func (p *participantSynchronizer) removeTrack(ssrc uint32) {
+	p.Lock()
+	defer p.Unlock()
+
+	if t := p.tracks[ssrc]; t != nil {
+		t.Lock()
```
boks1971: can we make this an API call to `TrackSynchronizer`. Would be good to keep locking inside that struct and just return the value.

milos-lk: updated

PR #878, opened 2026-04-07
Update to latest packet trailer feature format

chenosaurus:
- Update the packet trailer to match the latest protocol.  
- Update H264/H265 SEI NAL parser to expect the same format for frame metadata as packet trailer.


localparticipant.go
```diff
@@ -1103,6 +1105,20 @@ func (p *LocalParticipant) SendFile(filePath string, options StreamBytesOptions)
 	return &writer.Info, nil
 }
 
+func packetTrailerFeaturesFromOpts(opts *TrackPublicationOptions) []livekit.PacketTrailerFeature {
+	if opts == nil || (!opts.AttachUserTimestamp && !opts.AttachFrameId) {
+		return nil
+	}
```
boks1971, line 1111: small suggestion: instead of checking all the fields, to make this more scalable as more features get added, can just let the code go through and construct features and do a single check for `len(features) == 0` at the end.

packet_trailer.go
```diff
@@ -0,0 +1,190 @@
+package lksdk
```
boks1971, line 1: can we move this entire bit into `protocol` repo and use it both here and livekit server? One side will be adding and other side will be removing, but common place can potentially share all these constants at least.

Can do in a separate PR if that makes sense.

packet_trailer.go
```diff
@@ -0,0 +1,190 @@
...
+		pos += 2
+
+		if pos+length > tlvRegionLen {
+			break
+		}
+
+		valStart := tlvStart + pos
+
+		switch {
+		case tag == tagUserTimestamp && length == 8:
```
boks1971, line 134: should something like tag matching, but length not matching considered an error and the whole trailer rejected? Is there advantage to doing partial decode?

packet_trailer_h264_parser_test.go
```diff
@@ -0,0 +1,86 @@
...
+		userData := append(uuid[:], trailer...)
+
+		// payloadType = 5, payloadSize = len(userData)
+		nal = append(nal, 0x05, byte(len(userData)))
+		nal = append(nal, userData...)
+		return nal
+	}
+
+	t.Run("accepts matching UUID with timestamp only", func(t *testing.T) {
+		got, ok := parseH264SEIPacketTrailer(buildNAL(packetTrailerSEIUUID, wantMeta))
```
boks1971, line 25: Copilot flagged an issue in parsing the trailer about not considering `frameId`. I thought this SEI thing will have only timestamp and hence the Copilot comment is not relevant, but I see here mention of `frameId` also. So, should that routine consider `frameId` also like Copilot suggests?

packet_trailer_h264_parser_test.go
```diff
@@ -0,0 +1,86 @@
...
+		if !ok {
+			t.Fatalf("expected ok=true")
```
boks1971, line 27: We use the `require` package (https://pkg.go.dev/github.com/stretchr/testify/require) for tests. Would be good to switch all these to use `require`. Sorry if I missed mentioning this in a previous review.

readersampleprovider.go
```diff
@@ -343,10 +352,16 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
-					p.hasPendingUserTimestamp = true
+			if p.appendUserTimestamp || p.appendFrameId {
+				if meta, ok := parseH264SEIPacketTrailer(nalUnitData); ok {
+					if !p.appendUserTimestamp {
+						meta.UserTimestamp = 0
+					}
+					if !p.appendFrameId {
+						meta.FrameId = 0
+					}
+					p.pendingFrameMetadata = meta
```
boks1971, line 363: can this be a pointer type and pending can be checked for `pendingFrameMetadata != nil` instead of having an extra boolean to indicate that there is pending frame meta?

readersampleprovider.go
```diff
@@ -373,17 +388,10 @@ func (p *ReaderSampleProvider) NextSample(ctx context.Context) (media.Sample, er
...
-		// a zero timestamp.
-		if p.appendUserTimestamp {
-			ts := int64(0)
-			if p.hasPendingUserTimestamp {
-				ts = p.pendingUserTimestampUs
-				p.hasPendingUserTimestamp = false
-				p.pendingUserTimestampUs = 0
-			}
-			sample.Data = appendUserTimestampTrailer(sample.Data, ts)
+		if (p.appendUserTimestamp || p.appendFrameId) && p.hasPendingFrameMeta {
```
boks1971, line 391: can these be collapsed into a single setting? something like `appendPacketTrailer` so that it scales better if more trailers are added.

boks1971 (approved):
lgtm! some small comments

packet_trailer_h264_parser_test.go
chenosaurus: I will update this test.  We updated the parser to expect the content of the SEI frame to match the packet trailer format so it can support any of the metadata fields.

packet_trailer_h265_parser.go
```diff
@@ -0,0 +1,55 @@
+package lksdk
```
boks1971, line 15: can you please add the copyright file?

packet_trailer.go
boks1971, line 15: copyright file please.

boks1971 (approved):

Issue #870, opened 2026-03-27
Add H.264 video E2EE and data channel encryption for Go publishers


samthesloth:
## Summary

The Go SDK currently supports audio sample E2EE (`EncryptGCMAudioSample` from #573/#583) and PCM track encryption (#674), but there's no equivalent for **H.264 video frames** or **data channel messages**. This means Go publishers (screen sharing, server-side compositing, IoT cameras) can't participate in E2EE rooms.

## Use Case

We're building a remote desktop system where a Go runner captures the screen via FFmpeg, encodes to H.264, and publishes via the Go SDK. The JS viewer subscribes using the standard built-in `roomOpts.encryption` + `ExternalE2EEKeyProvider` + `e2ee-worker.js`. We need the Go publisher to produce encrypted frames that the JS FrameCryptor can decrypt.

## What We Built (fork)

We have a working implementation on a fork that adds:

### 1. H.264 video frame encryption (`encryption.go`)

`EncryptGCMH264Sample` / `EncryptGCMH264SampleCustomCipher` which match the JS FrameCryptor format exactly:

- Parses H.264 NALUs via `findNALUIndices` to find the first slice (IDR/non-IDR)
- Leaves everything before the slice + 2 bytes unencrypted (used as AES-GCM AAD)
- Samples with no slice NALUs (SPS/PPS only) pass through unmodified
- Encrypts payload with AES-128-GCM (random 12-byte IV)
- Applies RBSP escaping (`writeRBSP`) to prevent accidental H.264 start codes
- Frame format: `[header][ciphertext+tag][IV][IV_LENGTH][KID]`

Corresponding `DecryptGCMH264Sample` for Go-side decryption.

Uses the existing `DeriveKeyFromString` (PBKDF2, salt `"LKFrameEncryptionKey"`, 100k iterations) so the same shared passphrase works on both sides.

### 2. Data channel encryption (`data_cryptor.go`, `e2ee.go`)

- `DataCryptor` encrypts outgoing `DataPacket` values into `EncryptedPacket` protobuf wrappers
- Decrypts incoming `EncryptedPacket` messages back to their original types
- Supports: user messages, chat, RPC request/response/ack, stream headers/chunks/trailers
- `ExternalKeyProvider` with `SetKeyFromPassphrase` / `SetRawKey` and key index support
- `room.SetEncryption(&EncryptionOptions{KeyProvider: kp})` — integration point

### 3. Room/engine integration (`room.go`, `engine.go`)

[...]

PR #881, opened 2026-04-08
Add E2EE support for H.264/H.265 video and data channel encryption

samthesloth:
Regarding #870 

## Summary

- **H.264/H.265 video sample encryption**: AES-128-GCM with NALU-aware header preservation and RBSP escaping, compatible with the JS SDK `FrameCryptor`
- **Data channel encryption**: AES-128-GCM encryption of `DataPacket` values (user data, chat, RPC, streams), matching the JS SDK `DataCryptor`
- **Opt-in via ConnectOption**: `WithDataEncryption` follows the existing `With*` convention

## API

**Video track encryption** follows the existing `EncryptGCMAudioSample` pattern:

```go
// H.264
encrypted, err := lksdk.EncryptGCMH264Sample(sample, key, kid)
decrypted, err := lksdk.DecryptGCMH264Sample(encrypted, key, sifTrailer)

// H.265
encrypted, err := lksdk.EncryptGCMH265Sample(sample, key, kid)
decrypted, err := lksdk.DecryptGCMH265Sample(encrypted, key, sifTrailer)

// Cached cipher variants for performance
encrypted, err := lksdk.EncryptGCMH264SampleCustomCipher(sample, kid, block)
```

**Data channel encryption** is opt-in via a dedicated `ConnectOption`:

```go
kp := lksdk.NewExternalKeyProvider()
kp.SetKeyFromPassphrase("shared-secret", 0)

room.JoinWithToken(url, token,
    lksdk.WithDataEncryption(&lksdk.EncryptionOptions{KeyProvider: kp}),
)
```

## Design decisions

- **H.265 included** re reviewer feedback (good to have alongside H.264)
- **Data encryption as ConnectOption** re @cnderrauber's suggestion of a room option like `EnableDataEncryption`, implemented as `WithDataEncryption` to match existing `With*` conventions
- **`DataEncryptionKeyProvider any` in ConnectParams** uses `any` to avoid circular import between `signalling` and `lksdk` packages; type-asserted once in `JoinWithToken`

## Note on start code normalization

On Linux, FFmpeg can produce mixed 3-byte/4-byte H.264/H.265 start codes. The encrypt/decrypt functions handle both formats correctly, but decrypted output preserves the original start code format. If browser interop requires normalized 4-byte start codes, callers should normalize before encrypting. I have a `NormalizeStartCodes` utility ready as a follow-up if desired.

## Files changed

| File | Change |
|---|---|
| `e2ee.go` (new) | `KeyProvider` interface, `ExternalKeyProvider`, `DataCryptor` |
| `encryption.go` | H.264/H.265 encrypt/decrypt, NALU helpers, RBSP escaping |
| `encryption_test.go` | Tests + benchmarks for all new encryption |
| `engine.go` | `DataCryptor` field, encrypt/decrypt hooks in data channel path |
| `room.go` | `WithDataEncryption` ConnectOption |
| `signalling/interfaces.go` | `DataEncryptionKeyProvider` field on ConnectParams |

## Test plan

- [x] H.264 encrypt/decrypt round-trip (IDR + non-IDR slices)
- [x] H.264 SPS/PPS-only frames pass through unmodified
- [x] H.265 encrypt/decrypt round-trip (all 16 VCL slice types)
- [x] H.265 reserved NALU types pass through
- [x] DataCryptor round-trip (UserPacket -> encrypt -> decrypt -> verify payload)
- [x] Edge cases: nil cipher, too-short samples, wrong key length, empty passphrase
- [x] `go vet` clean


e2ee.go
```diff
@@ -0,0 +1,260 @@
+package lksdk
```
cnderrauber, line 15: We can create a new folder for e2ee like js-sdk to avoid `any` type in ConnectParams, and put all e2ee related code in a dedicated package. But don't move the frame encryptor/decryptor now for api compatibility, we can move it later when bump the sdk version.
e2ee
&emsp;|--types
&emsp;&emsp;|--interface.go <- put `KeyProvider` interface here
&emsp;|--keyprovider.go
&emsp;|--datacryptor.go

cnderrauber (commented):
The video encryptor/decryptor is sample level only that means user needs to call func for each frame when send/recv. Need to add a track level option and integrate media e2ee to `LocalTrack` to make it easy to use.

samthesloth:
I have made a few changes to better implement E2EE especially with @cnderrauber comments :

- Create new /e2ee/ package with `KeyProvider`, `ExternalKeyProvider`, `DataCryptor`, and types/interface.go (matches JS SDK's src/e2ee/ layout).
- Type `ConnectParams.DataEncryptionKeyProvider` as `e2eetypes.KeyProvider` instead of `any`.
- Leave `EncryptGCSample`/ `DecryptGCMSample` in encryption.go untouched for backward compatibility.
- Add `LocalTrack.WithFrameEncryptor(enc)` so `WriteSample` encrypts frames automatically before packetization.
- Provide `lksdk.NewFrameEncryptor(kp, codec)` / `NewFrameDecryptor(...)` with a Codec enum (H264/H265/Opus), matching the JS SDK `FrameCryptor` shape.
- Fix SIF-trailer check in `GCMFrameDecryptor.DecryptFrame` (was looking up a key by the last byte before checking if the frame was server-injected, which failed with "no key at index N" for SIF frames).
- Add round-trip tests for mixed 3-byte/4-byte H.264/H.265 Annex-B start codes

Edit: Please let me know if I should add ratchet support and/or room-level encryption with E2EE similar to JS (currently, I do it per track as requested and more in line with the audio encryption implementation).

e2ee.go
samthesloth: Hello! Thank you so much for the review! I have just made a new commit.

e2ee/frameencryptor.go
```diff
@@ -0,0 +1,160 @@
...
+	e := &GCMFrameEncryptor{
+		keyProvider: keyProvider,
+		encryptFn:   encryptFn,
+	}
+	e.state.Store(&cipherState{block: block, keyIndex: idx})
+	return e, nil
+}
+
+// EncryptFrame encrypts a complete media frame. Called from the hot path
+// (LocalTrack.WriteSample) at hundreds of frames/sec.
```
cnderrauber: don't need to mention call path in the comment, and also hundreds fps is abnormal.

encryption.go
```diff
@@ -245,3 +257,315 @@ func EncryptGCMAudioSampleCustomCipher(sample []byte, kid uint8, cipherBlock cip
 	return newData, nil
 
 }
+
```
cnderrauber: can also move these encrypt/decrypt function to e2ee package

e2ee/trackdecryptor.go
```diff
@@ -0,0 +1,110 @@
...
+// into a complete frame, and decrypts the frame. Returns (nil, nil) for
+// server-injected frames (SIF) that should be dropped. Returns io.EOF
+// when the track is closed.
+func (td *TrackDecryptor) ReadSample() (*media.Sample, error) {
+	for {
+		// Feed packets into the samplebuilder until a frame is ready.
+		sample := td.sb.Pop()
+		if sample != nil {
+			decrypted, err := td.decryptor.DecryptFrame(sample.Data)
+			if err != nil {
```
cnderrauber, line 80: The error is swallowed silently, we should either throw the error to caller, or log a warn message (with a counter to avoid flood).

e2ee/frameencryptor.go
```diff
@@ -0,0 +1,160 @@
...
+	block, err := d.getCachedBlock(kid)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.decryptFn(payload, d.sifTrailer, block)
+}
+
+func (d *GCMFrameDecryptor) getCachedBlock(kid uint32) (cipher.Block, error) {
+	if v, ok := d.cache.Load(kid); ok {
```
cnderrauber: it could hit a invalid cache when the key is refreshed in keyProvider, datacryptor have same problem.

cnderrauber (commented):
Thank you for the working!! It would be better to have a example `videotracke2ee` for developer and a e2ee test case in `integration_test.go` .
cc @boks1971 @lukasIO to get more views on this.

engine.go
```diff
@@ -628,6 +632,17 @@ func (e *RTCEngine) handleDataPacket(msg webrtc.DataChannelMessage) {
...
+
+	// Decrypt if data channel E2EE is enabled and this is an encrypted packet.
+	if ep, ok := packet.Value.(*livekit.DataPacket_EncryptedPacket); ok && e.dataCryptor != nil {
+		payload, err := e.dataCryptor.Decrypt(ep.EncryptedPacket)
+		if err != nil {
+			e.log.Warnw("data decryption failed, dropping packet", err)
+			return
+		}
+		decryptedPayloadToDataPacketValue(packet, payload)
+	}
```
boks1971, line 648: maybe log an error and return if an encrypted packet is received, but there is not decryptor.

e2ee/datacryptor.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"github.com/livekit/protocol/livekit"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/livekit/server-sdk-go/v2/e2ee/types"
+)
+
+const (
+	ivLength = types.IVLength
+)
```
boks1971: can types.IVLength used directly wherever it is used?

integration_test.go
```diff
@@ -563,3 +564,93 @@ func TestSimulcastCodec(t *testing.T) {
 	}
 
 }
+
+// TestE2EE_AudioRoundTrip verifies that publishing an encrypted audio track
+// flows through the SFU: the track is received by the subscriber, the
+// encryption metadata survives signaling, and the e2ee API (key provider,
+// frame decryptor, SIF trailer) composes without error. It also exercises
+// key rotation via SetRawKey at the same key index, which guards against the
```
boks1971: don't need to mention the bug. For somebody reading this, there is no context of what the bug is. As long as bug is fixed, don't need to mention it.

samthesloth:
Thanks for the review! I pushed a follow-up addressing all comments:

- `frameencryptor.go`: Dropped the hot-path comment (and edited similar phrasing elsewhere in e2ee/)
- `encryption.go`: Moved the H.264/H.265 functions + video helpers into new files under e2ee/ (`codec_h264.go`, `codec_h265.go`, `codec_video.go`). I kept audio where it was to avoid breaking the existing lksdk.* public API
- `trackdecryptor.go`: `ReadSample` now returns the decrypt error instead of silently dropping the frame
- `getCachedBlock` and `getCipherBlock` : Cache now stores `(keyBytes, block)` and compares bytes against `keyProvider.GetKey` on every lookup, rebuilding on mismatch. No interface changes. Added regression tests in `frameencryptor_test.go` and `datacryptor_test.go` 
- New examples/videotracke2ee showing H.264 e2ee publish/subscribe via `NewFrameEncryptor` / `NewFrameDecryptor` +
  `TrackDecryptor`.
- New `TestE2EE_H264RoundTrip` in `integration_test.go` — publishes a GCM-flagged H.264 track, asserts metadata propagates, exercises key rotation via `SetKeyFromPassphrase` at the same index.

e2ee/types/constants.go
```diff
@@ -0,0 +1,18 @@
+package types
```
boks1971, line 15: can you please add the copyright notice to all the new files? You can copy it from an existing file.

e2ee/codec_h265_test.go
```diff
@@ -0,0 +1,142 @@
...
+	frame = append(frame, 0x00, 0x00, 0x00, 0x01) // 4-byte
+	frame = append(frame, spsHeader...)
+	frame = append(frame, 0xCC, 0xDD)
+	frame = append(frame, 0x00, 0x00, 0x01) // 3-byte
+	frame = append(frame, idrHeader...)
+	frame = append(frame, 0x11, 0x22, 0x33, 0x44, 0x55)
+
+	encrypted, err := e2ee.EncryptGCMH265Sample(frame, key, 0)
+	require.NoError(t, err)
+	require.NotEqual(t, frame, encrypted)
```
boks1971, line 127: is it possible to add checks for non-encrypted parts (header/sps/pps etc go through)?

boks1971 (approved):
great stuff @samthesloth . lgtm with some small comments.

Would be good to get a nod from @lukasIO .

I think the room level stuff can be a separate PR. This is already very big. Would like to get @lukasIO 's thoughts on room level stuff.

Can you please rebase and resolve the conflicts?

samthesloth:
@boks1971 Pushed commit with requested changes. Just to make sure, you're asking me to take out the data channel encryption stuff I added at the room level? If so, I can make a follow-up commit removing that before rebasing :)

boks1971:
Thank you so much.

No, not take out stuff.

I forget the details now, but if JS SDK has a room level setting for tracks, that's what I meant in a different PR.

lukasIO:
>  but if JS SDK has a room level setting for tracks

JS SDK has a room level setting that would (by default) ensure encryption on all outgoing tracks. 
For incoming tracks both encrypted and unencrypted are supported when the decryptor is set up accordingly. 

Datachannel encryption is enabled by default via the same setting, but as mentioned in the linked issue (#870) I think it makes sense for this to be a separate setting here in order to maximise compatibility. 

PR generally looks good to me as long as folks here are happy with the API design for this. 
cc @anunaym14 as he had implemented the original e2ee code in this repo

cnderrauber (approved):

PR #882, opened 2026-04-08
fix bug where projects with "localhost" in the URL are misrouted

andrewnitu:
The localhost check in getAgentsURL (line 157) uses strings.Contains to detect local development URLs, but this also matches project subdomains that happen to contain "localhost" as a substring (e.g. air-localhost-ist1jyfs).
When this happens, the regex replacement that rewrites the project subdomain to agents.livekit.cloud is skipped entirely. Requests to /build and other agent endpoints end up hitting the *.livekit.cloud wildcard ingress on the core cluster instead of the agents service, and the cloud-controller catch-all returns "OK" — which fails to decode as a valid response.

### DIFF
--- a/pkg/cloudagents/client.go
+++ b/pkg/cloudagents/client.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"io/fs"
 	"net/http"
+	"net/url"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/cloudagents/client.go
```diff
@@ -169,13 +181,25 @@ func (c *Client) getAgentsURL(serverRegion string) string {
...
+	pattern := `^https://[a-zA-Z0-9\-]+\.`
+	re := regexp.MustCompile(pattern)
+	if serverRegion != "" {
+		serverRegion = fmt.Sprintf("%s.", serverRegion)
+	}
+	return re.ReplaceAllString(agentsURL, fmt.Sprintf("https://%sagents.", serverRegion))
+}
+
+func isLocalURL(rawURL string) bool {
+	u, err := url.Parse(rawURL)
```
real-danm, line 199: can you add some tests for this? make sure we can support the project names with 'localhost' and a few other formats?

yoonsio (commented):
Would appreciate it if you could provide some context.
Do we have support for localhost projects?

andrewnitu:
@yoonsio a customer has a project name with "localhost" in it and the builds totally fail for them. see https://live-kit.slack.com/archives/C08HHUXHAKA/p1775494175027349 for details

tbh i don't know why this localhost thing is here, my guess is if you're testing multiple services on your laptop? Trying to not regress that workflow.

pkg/cloudagents/client.go
andrewnitu: done

real-danm (approved):

PR #885, opened 2026-04-13
bump version to 2.16.2

andrewnitu:


anunaym14 (approved):

anunaym14:
My changes are good to go :)

yoonsio (approved):
Does this introduce any breaking changes? For example, if we use this server-sdk-go v2.16.2 in cloud-agents, can the current version of livekit-cli v2.16.0 be used? If yes, LGTM! Otherwise, we should consider bumping the minor version and enforce min cli version.

PR #886, opened 2026-04-17
TEL-522: JoinWithToken accepts a context to be cancelled or timed out early

hechen-eng:
Even though `CreateSIPParticipant` timed out much earlier and agent left, `JoinWithToken` continued to retry and eventually succeeded and the sip participant joined a room with no agent in it. We should stop retrying if `CreateSIPParticipant` already timed out hence the passed in context.

### DIFF
--- a/room.go
+++ b/room.go
@@ -111,6 +111,15 @@ func WithConnectTimeout(timeout time.Duration) ConnectOption {
 	}
 }
 
+// WithContext sets a context for the JoinWithToken call.
+// When provided, the region discovery retry loop and backoff sleeps will
+// respect context cancellation/deadline.
+func WithContext(ctx context.Context) ConnectOption {
+	return func(p *signalling.ConnectParams) {
+		p.Context = ctx
+	}
+}
+
 // WithAutoSubscribe sets whether the participant should automatically subscribe to tracks.
 // Default is true.
 func WithAutoSubscribe(val bool) ConnectOption {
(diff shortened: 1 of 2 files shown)


biglittlebigben:
Should we consider making a breaking API change here? putting the context as the first parameter in the function call seems a lot more idiomatic (and not having it an oversight)

hechen-eng:
> Should we consider making a breaking API change here? putting the context as the first parameter in the function call seems a lot more idiomatic (and not having it an oversight)

I don't see why not 🤔

alexlivekit (approved):
lgtm, but my knowledge here is not thorough.

+1 To the suggestion of just break the API and force the context.

biglittlebigben:
@boks1971 What do you think? How bad would it be to break the Join/JoinWithToken APIs to require a context parameter?

boks1971:
> @boks1971 What do you think? How bad would it be to break the Join/JoinWithToken APIs to require a context parameter?

Not too bad I think. Users would have to include a `context.Backgroud()` (they should not give `nil` though or we need to check for nil).

But, my vote would be for extending the API to something like https://pkg.go.dev/net/http#NewRequestWithContext and make a `JoinWIthContext` and `JoinWithContextAndToken` so that whoever needs it can use it. Older users can take the time to switch. It is closer to what He has done, but a API level function and it is clear as it does the idiomatic first argument context.

biglittlebigben:
> > @boks1971 What do you think? How bad would it be to break the Join/JoinWithToken APIs to require a context parameter?
> 
> Not too bad I think. Users would have to include a `context.Backgroud()` (they should not give `nil` though or we need to check for nil).
> 
> But, my vote would be for extending the API to something like https://pkg.go.dev/net/http#NewRequestWithContext and make a `JoinWIthContext` and `JoinWithContextAndToken` so that whoever needs it can use it. Older users can take the time to switch. It is closer to what He has done, but a API level function and it is clear as it does the idiomatic first argument context.

This would work as well.

hechen-eng:
@biglittlebigben @boks1971 done!

PR #890, opened 2026-04-23
sync engine

frostbyte73:
New Synchronizer based on Chrome's `RtpToNtpEstimator`, `StreamSynchronization`, and `RtpStreamsSynchronizer`

- `NtpEstimator`: Per-track linear regression (OLS over last 20 RTCP Sender Reports) mapping RTP timestamps to NTP wall-clock time.
- `ParticipantSync` — Compares NTP-mapped times across one participant's audio and video tracks to compute A/V lip sync adjustments, with time-based slew rate limiting (5ms/s) and a 5ms deadband.
- `SessionTimeline` — Maps each participant's independent NTP clock domain onto a shared recording timeline using one-way delay estimation to normalize out clock offsets between participants.
- `SyncEngine` — Orchestrator that wires the three layers together, handles wall-clock fallback before SRs arrive, manages the transition to NTP-grounded PTS, and enforces monotonicity and drain. Implements the Sync/TrackSync interfaces so AppWriter is agnostic to which engine is active.


boks1971 (approved):
not able to read it all, looks good from what I could read

milos-lk:
started looking into this today - I need to set aside couple of hours tomorrow to review it in depth.

pkg/synchronizer/syncengine.go
```diff
@@ -0,0 +1,494 @@
...
+	// Step 5: Apply ParticipantSync A/V adjustment.
+	st.engine.timeline.mu.RLock()
+	if pc, ok := st.engine.timeline.participants[st.identity]; ok {
+		adj := pc.participantSync.GetAdjustment(st.track.ID())
+		pts += adj
+	}
+	st.engine.timeline.mu.RUnlock()
+
+	// Step 6: Enforce monotonicity.
+	if pts < st.lastPTSAdjusted+time.Millisecond && st.lastPTSAdjusted > 0 {
```
milos-lk: does this force e.g same frame packets to have 1ms advance? if yes - that could break muxers.

pkg/synchronizer/syncengine.go
```diff
@@ -0,0 +1,494 @@
...
+		st.mu.Lock()
+		if st.lastPTSAdjusted > maxPTS {
+			maxPTS = st.lastPTSAdjusted
+		}
+		st.mu.Unlock()
+	}
+
+	startedAt := e.startedAt.Load()
+	if startedAt > 0 {
+		e.endedAt.Store(startedAt + int64(maxPTS))
```
milos-lk, line 275: I think this brings back bug fixed here: https://github.com/livekit/server-sdk-go/pull/868/changes

milos-lk:
The old sync drops packets older than oldPacketThreshold. The new sync has no equivalent. Very old packets coming in a burst could fill in gst queues and slow down burst recovery. Is there any reason not to have it inside the new sync?

milos-lk:
Similar to the note above - new sync doesn't integrate pipeline time feedback (existing `normalizePTSToMediaPipelineTimeline`) so that workaround we built looks like not used in the new solution? media running time is set though..

milos-lk:
Synchronization related bugs were the hardest class of bugs I have seen for egress in the previous period. I think we need to have good logging around this also in the new logic. Adjustments performed, stats etc. The existing syncrhonizer code has decent amount of logs which came to be through the pain of debugging sync issues.

pkg/synchronizer/syncengine.go
```diff
@@ -0,0 +1,494 @@
...
+	if st.lastTS == ts && st.lastPTS > 0 {
+		return st.lastPTS
+	}
+
+	// Wall-clock elapsed since this track started, plus session offset
+	wallElapsed := pkt.ReceivedAt.Sub(st.startTime) + st.sessionOffset
+
+	// If we have a previous timestamp, use RTP delta for more precision.
+	if st.lastPTS > 0 {
+		rtpDelta := ts - st.lastTS
```
milos-lk: is this wraparound safe? It will probably be rejected because of `wallClockSanityThreshold` but that wallcklock time will be returned

pkg/synchronizer/syncengine.go
```diff
@@ -0,0 +1,494 @@
...
+			st.ntpTransitioned = true
+		}
+		pts = ntpPTS
+	}
+
+	// Step 4: Apply transition slew (absorb gradually toward zero, time-based).
+	if st.transitionSlew != 0 {
+		pts += st.transitionSlew
+
+		now := pkt.ReceivedAt
```
milos-lk: transition slew is using received time - this might slow down slew on bursts as elapsed will stay very low value (close to 0) for these packets received on burst. Maybe using wallckock time instead?
or maybe it could be driven on session elapsed time, same as ParticipantSync.updateAdjustments is already driven. The slew state moves into a periodic tick instead of being computed inline in GetPTS.. not sure what's better/less complex

milos-lk:
Maybe it's worth calling out that some of the options being in the current sync logic aren't supported - like disabling PTS adjustments -  I guess we can use existing synchronizer for cases where this is needed

milos-lk:
Pretty neat idea of using both NtpEsitmator and OWD estimator to map everything through session timeline - I like that 👍

frostbyte73:
> Maybe it's worth calling out that some of the options being in the current sync logic aren't supported - like disabling PTS adjustments - I guess we can use existing synchronizer for cases where this is needed

My hope was that if we just ripped all the logic out of chrome, we wouldn't need those options. Doing this in parallel with the sdk compositing so we have a good way to test/measure other than just deploy to canary and hope

milos-lk:
> > Maybe it's worth calling out that some of the options being in the current sync logic aren't supported - like disabling PTS adjustments - I guess we can use existing synchronizer for cases where this is needed
> 
> My hope was that if we just ripped all the logic out of chrome, we wouldn't need those options. Doing this in parallel with the sdk compositing so we have a good way to test/measure other than just deploy to canary and hope

👍 definitively nice for side by side comparison. I am afraid though we might need it, but let's see 🤞 . A lot of straightening for chrome is happening on DSP level, inside libwebrtc with NetEq which is what e.g fixes audio timeline with changing tempo and not touching PTS (more complex and probably robust logic than what I built with pitch audio tempo control.).

pkg/synchronizer/syncengine.go
frostbyte73: ah, yeah chrome deals with decoded frames so this doesn't work for us. Added the same pts early return

pkg/synchronizer/syncengine.go
frostbyte73: yeah subtraction is wraparound safe

frostbyte73:
Integration tests all passing https://github.com/livekit/egress/pull/1198

boks1971:
Sorry @frostbyte73 . Backlogged on this. Will try to read through this today.

pkg/synchronizer/ntpestimator.go
```diff
@@ -0,0 +1,282 @@
...
+	var sumDxDx, sumDxDy float64
+	e.iterSamples(func(s srSample) {
+		dx := float64(s.unwrappedRTP) - mX
+		dy := float64(s.ntpNanos) - mY
+		sumDxDx += dx * dx
+		sumDxDy += dx * dy
+	})
+
+	if sumDxDx == 0 {
+		// Degenerate case: all RTP timestamps identical.
```
milos-lk, line 219: when running locally - audio is pretty broken for e.g participant recordings - one of the observation from logs (not sure if that's the only thing breaking it):
```
2026-04-26T12:30:59.452Z	WARN	egress	synchronizer/syncengine.go:457	NTP PTS exceeds trust threshold, clamping to wall clock	{"nodeID": "NE_CNvpcyxbFKj6", "handlerID": "EGH_iVneuQXe2GaT", "clusterID": "", "egressID": "EG_xYc6NL7a5U5q", "trackID": "TR_AMuhJq2oNDupA4", "kind": "audio", "syncEngine": true, "ntpPTS": "-493668h30m38.244798215s", "wallPTS": "21.167979203s", "diff": "-493668h30m59.412777418s"}
```

was wondering how the `ntpPTS` could have that value - could it be that this returns early here and we declare `NtpEstimator` ready on the line just after calling this function? Could it be 2 compound RTCP packets with the same RTP timestamps causing this?

milos-lk: just asked claude - it confirmed that possibility ^^^ and added this on top:
```
NTP jump detection accumulates without bound — once the regression fixes, every packet triggers 
  a new correction because the detection doesn't account for the correction it already applied       
   
  The fix: lastNtpPTS should be set to the corrected PTS (after ntpCorrection is applied), 
  not the raw NTP PTS. That way the next packet's expected PTS accounts for the correction, and no 
  new jump is detected. Currently the code sets lastNtpPTS = pts before the ntpCorrection decay is   
  applied, creating the feedback loop.
```

pkg/synchronizer/participantsync.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+// SetTrackEstimator registers or updates the NtpEstimator for a given track.
+func (ps *ParticipantSync) SetTrackEstimator(trackID string, mediaType MediaType, estimator *NtpEstimator) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	if existing, ok := ps.tracks[trackID]; ok {
+		existing.estimator = estimator
```
boks1971: can the estimator change mid-stream? how does that impact sync calculations?

pkg/synchronizer/participantsync.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+)
+
+// MediaType distinguishes audio and video tracks for A/V sync purposes.
+type MediaType int
+
+const (
+	MediaTypeAudio MediaType = iota
+	MediaTypeVideo
+)
```
boks1971: Just making a note for myself to check on how this is used. Main reason is screen share + screen share audio, i. e. are they synced? Should this be using `TrackSource` from protocol repo?

pkg/synchronizer/ntpestimator.go
```diff
@@ -0,0 +1,282 @@
...
+	unwrapped := e.unwrapRTP(rtpTimestamp)
+
+	// Outlier rejection: if we already have a valid regression, check whether
+	// this new sample deviates from the prediction by more than 3 standard
+	// deviations.
+	if e.ready && e.residStd > 0 {
+		predicted := e.slopeNanos*(float64(unwrapped)-e.meanX) + e.meanY
+		residual := math.Abs(float64(ntpNanos) - predicted)
+		if residual > outlierThresholdStdDevs*e.residStd {
+			// Reject this sample as an outlier.
```
boks1971: probably worth logging initially to check how far off things get.

pkg/synchronizer/ntpestimator.go
```diff
@@ -0,0 +1,282 @@
...
+		ntpNanos:     ntpNanos,
+		receivedAt:   receivedAt,
+	}
+	e.sampleHead = (e.sampleHead + 1) % maxSRSamples
+	if e.sampleLen < maxSRSamples {
+		e.sampleLen++
+	}
+
+	// Recompute regression if we have enough samples.
+	if e.sampleLen >= 2 {
```
boks1971: should we have a min samples here? Wondering if two samples (assume one of it is an outlier) skew things resulting in later normal samples looking like outliers?

On the flip side, the readiness will get delayed wait for more reports. SFU currently does once in three seconds. So, it will add at that rate the more we wait.

pkg/synchronizer/ntpestimator.go
boks1971: pion will broadcast the RTCP packets to all contained SSRCs. I think there was a path which was causing the track synchroniser to get same RTCP sender reports twice. Had added this in the legacy track synchroniser to filter out duplicates - https://github.com/livekit/server-sdk-go/pull/820/changes

pkg/synchronizer/ntpestimator_test.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func TestNtpEstimator_SlidingWindow(t *testing.T) {
+	const clockRate = 90000
+	e := NewNtpEstimator(clockRate)
+
+	baseTime := time.Date(2025, 4, 1, 0, 0, 0, 0, time.UTC)
+
+	// Feed 25 SRs (exceeds window of 20)
+	for i := 0; i < 25; i++ {
```
boks1971, line 183: :nit: make this `maxSRSamples + 5` so that this is okay even if we change max, comment also needs to change

pkg/synchronizer/sessiontimeline.go
```diff
@@ -0,0 +1,256 @@
...
+	return pc
+}
+
+// GetOrAddParticipant returns the ParticipantClock for the given identity,
+// creating one if it doesn't exist. This is safe for concurrent use.
+func (st *SessionTimeline) GetOrAddParticipant(identity string) *ParticipantClock {
+	st.mu.Lock()
+	defer st.mu.Unlock()
+
+	if pc, ok := st.participants[identity]; ok {
```
boks1971: is this participant identity or SID? If there is a reconnect, identity would be the same. That might re-use the estimator which could be a different NTP base.

pkg/synchronizer/sessiontimeline.go
```diff
@@ -0,0 +1,256 @@
...
+
+// RemoveParticipant removes the participant with the given identity.
+func (st *SessionTimeline) RemoveParticipant(identity string) {
+	st.mu.Lock()
+	defer st.mu.Unlock()
+	delete(st.participants, identity)
+}
+
+// OnSenderReport processes an RTCP sender report for a participant's track.
+// It updates the NTP estimator, OWD estimator, and records the NTP epoch.
```
boks1971: Looks like this comment should be for the next function.

pkg/synchronizer/sessiontimeline.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+}
+
+// SetSessionStart sets the session start time (wall-clock time when the first
+// packet of any track arrived at the receiver).
+func (st *SessionTimeline) SetSessionStart(t time.Time) {
+	st.mu.Lock()
+	defer st.mu.Unlock()
+	st.sessionStart = t
+	st.hasStart = true
```
boks1971, line 66: Looks like this is not used.

Should `GetSessionPTS` be checking this and return `errNotReady`?

pkg/synchronizer/syncengine.go
```diff
@@ -0,0 +1,655 @@
...
+	}
+}
+
+// SyncEngine orchestrates NtpEstimator, ParticipantSync, and SessionTimeline
+// to provide cross-participant alignment and per-participant A/V lip sync.
+// It implements the Sync interface.
+type SyncEngine struct {
+	mu       sync.Mutex
+	timeline *SessionTimeline
+	tracks   map[uint32]*syncEngineTrack // keyed by SSRC
```
boks1971, line 91: are we tracking SSRC changes?

The SSRC changes on SFU down track if the publisher migrates in (down track will change from  remote track -> local track subscription) or migrates out (the reverse). That causes the down track to be closed and re-opened on SFU side. But, it is all internal to SFU, i. e. no negotiation. But, the SSRC changes though as it happens in pion and we do not have a way to set an SSRC (maybe need to add one). So, just want to make sure that trackID remaining same, but SSRC changing would not cause any issues.

Will also be an issue if the subscriber migrates. But, Go SDK does not support migration. So, it will be a full reconnect of egress participant. That should work ok.

pkg/synchronizer/participantsync.go
boks1971: After reading through it all, looks like `MediaType` is not really use. Would be good to clean up if not needed.

boks1971 (approved):
Good stuff @frostbyte73 ! Left some comments.

I am not able to check on every bit as there are a lot of scars from older battles. I think @milos-lk caught some of those and you have included it all. Guess, this will require good chunk of testing + good amount of soak in `staging`.

pkg/synchronizer/ntpestimator.go
frostbyte73: Increased to 4 SRs, 2 had problems in testing

pkg/synchronizer/ntpestimator.go
frostbyte73: yeah it was the duplicate SR issue, has been fixed

pkg/synchronizer/syncengine.go
frostbyte73: yeah, it should be handled properly. It'll be a full state reset for the track though

boks1971: Need to check if we can add something to pion to set the SSRC.

pkg/synchronizer/sessiontimeline.go
frostbyte73: replaced with SID

pkg/synchronizer/participantsync.go
frostbyte73: not anymore, removing

milos-lk (approved):
We might need to update drift value for the case when we don't touch audio PTS as we discussed yesterday - but other than that 👍

PR #892, opened 2026-04-24
fix(PCMLocalTrack): do not spawn goroutine until track is bound

anunaym14:
Also switch `closed` state to `fuse`

### DIFF
--- a/pkg/media/pcmlocaltrack.go
+++ b/pkg/media/pcmlocaltrack.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/frostbyte73/core"
 	"github.com/gammazero/deque"
 	"github.com/google/uuid"
 	"github.com/livekit/media-sdk"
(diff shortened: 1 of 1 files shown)


boks1971 (approved):

paulwe (commented):
is this testable?

anunaym14:
> is this testable?

do you mean PCMLocalTrack in general? or this particular code change?

In general - i think the individual writer bits can be tested in the media-sdk repo but overall, it might not be easy to do since opus is a lossy codec. 

For this particular change - i don't think it needs testing. It is just preventing leaking a goroutine when a track is created but never published.

biglittlebigben:
> > is this testable?
> 
> do you mean PCMLocalTrack in general? or this particular code change?
> 
> In general - i think the individual writer bits can be tested in the media-sdk repo but overall, it might not be easy to do since opus is a lossy codec.
> 
> For this particular change - i don't think it needs testing. It is just preventing leaking a goroutine when a track is created but never published.

Role of a test here would be:
- Make sure that the bug is actually fixed
- Make sue this doesn't get reintroduced by accident
- maybe not here but generally: document expected behavior

PR #896, opened 2026-04-27
Hosted agents:  first-cut adding agentName to lk commands

shawnyang-coder:


go.mod
```diff
@@ -149,4 +149,6 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+replace github.com/livekit/protocol => ../protocol
```
shawnyang-coder, line 152: This replace line is only needed to get my local build to pass. I will cut a release on protocol and change to use that release before merging this pr.

PR #898, opened 2026-04-29
[Minor fixes]  correct version comparison, nil-safe speaker update, and RLock/RUnlock mismatch

siddimore:
### Why

Three independent issues found during a code walk through:

-  compareVersions produced wrong results for double-digit version components — the > branch used string comparison ("10" > "9" is false lexicographically) even though the < branch already used the correctly parsed integers.
- OnSpeakersChanged used reflect to detect a nil participant — assigning a *RemoteParticipant directly into a Participant interface before checking nil creates a non-nil interface wrapping a nil pointer, so a plain == nil check silently fails. Using reflect is both unusual and unnecessary here.
- getPublication held a write lock for a read-only operation — the lock acquisition was changed to RLock but defer p.lock.Unlock() was left behind, which would panic at runtime with a mismatched unlock.

### What changed

File	Change
utils.go	Use parsed integer p1 > p2 instead of string parts1[i] > parts2[i]
room.go	Check *RemoteParticipant nil before boxing into interface; remove reflect import
participant.go	Fix defer p.lock.Unlock() → defer p.lock.RUnlock() to match the RLock
utils_test.go	New table-driven tests for compareVersions, including the double-digit regression case

### How tested

- Added TestCompareVersions in utils_test.go covering equal, greater, less, double-digit components ("10.0.0" vs "9.0.0"), and differing length versions — all pass.
- The room.go and participant.go changes are mechanical and covered by existing room/participant tests.

### DIFF
--- a/participant.go
+++ b/participant.go
@@ -266,8 +266,8 @@ func (p *baseParticipant) addPublication(publication TrackPublication) {
 }
 
 func (p *baseParticipant) getPublication(sid string) TrackPublication {
-	p.lock.Lock()
-	defer p.lock.Unlock()
+	p.lock.RLock()
+	defer p.lock.RUnlock()
 
 	track, ok := p.tracks.Load(sid)
 	if !ok {
--- a/room.go
+++ b/room.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"fmt"
 	"maps"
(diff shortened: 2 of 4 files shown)


cnderrauber (approved):
Looks good!

PR #899, opened 2026-04-29
[Minor] logging fixes

siddimore:
### Why

Several silent failure paths made it a bit harder to debug and understand the reason while debugging

### Changes
- `engine.go` — reconnect backoff: Replaced the dead break (delay > maxReconnectInterval) with a min() clamp. With quadratic growth and maxReconnectCount=10, the break was unreachable (max delay is 24.3s vs 60s cap). If maxReconnectCount were ever increased, the break would exit the loop early and skip the OnDisconnected call, leaving the room in a zombie state.
- `engine.go` — data packet parse error: Was silently returning on error. Now logs Warnw with the error
- `engine.go` — unknown DataChannel label: Was silently returning. Now logs Warnw with the label 
- room.go — reconnect state transitions: OnRestarting, OnResuming, and OnResumed now log Infow at the room level, making the reconnect lifecycle traceable without relying solely on engine-level logs.
- `remoteparticipant.go` — track subscription timeout: The 5s metadata wait goroutine now logs Warnw with trackSID
- `localparticipant.go` — wrong logger instance: updateSubscriptionPermissionLocked was using the global logger instead of p.log

### Testing
go build . passes cleanly

### DIFF
--- a/room.go
+++ b/room.go
@@ -871,6 +871,7 @@ func (r *Room) OnDisconnected(reason DisconnectionReason) {
 }
 
 func (r *Room) OnRestarting() {
+	r.log.Infow("room connection restarting")
 	r.setConnectionState(ConnectionStateReconnecting)
 	r.callback.OnReconnecting()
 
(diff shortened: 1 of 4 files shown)


room.go
```diff
@@ -871,6 +871,7 @@ func (r *Room) OnDisconnected(reason DisconnectionReason) {
 }
 
 func (r *Room) OnRestarting() {
+	r.log.Infow("room connection restarting")
```
cnderrauber, line 874: there are engine logs for these events

PR #905, opened 2026-05-13
Add option to publish tracks with join request and enable single pc mode

cnderrauber:
Adds WithSinglePeerConnection / WithTrack / WithSimulcastTrack connect options that bundle the AddTrackRequest and publisher offer into the JoinRequest, saving time on connection/publishing.

Benchmark with 100ms rtt, 10 rounds
  | case                             | min    | p90    | avg    | max    | saved (ms / RTT) |
|----------------------------------|--------|--------|--------|--------|------------------|
| dual-pc post-publish / audio     | 1.049s | 1.205s | 1.136s | 1.216s | — (baseline)     |
| single-pc post-publish / audio   | 967ms  | 1.031s | 986ms  | 1.056s | 150 ms / 1.4 RTT |
| single-pc pre-publish / audio    | 855ms  | 876ms  | 866ms  | 880ms  | 270 ms / 2.5 RTT |
| dual-pc post-publish / video     | 1.073s | 1.277s | 1.170s | 1.282s | — (baseline)     |
| single-pc post-publish / video   | 954ms  | 1.064s | 1.019s | 1.133s | 151 ms / 1.4 RTT |
| single-pc pre-publish / video    | 879ms  | 977ms  | 931ms  | 978ms  | 239 ms / 2.2 RTT |

### DIFF
--- a/signalling/interfaces.go
+++ b/signalling/interfaces.go
@@ -100,6 +100,10 @@ type ConnectParams struct {
 	// DataEncryptionKeyProvider enables data channel E2EE when set.
 	DataEncryptionKeyProvider e2eetypes.KeyProvider
 
+	// UseSinglePeerConnection enables single peer connection mode in which the
+	// publisher PC is reused for both publishing and receiving media.
+	UseSinglePeerConnection bool
+
 	// internal use
 	Codecs []webrtc.RTPCodecParameters
 
--- a/engine.go
+++ b/engine.go
@@ -94,6 +94,12 @@ const (
 	maxReconnectCount        = 10
 	initialReconnectInterval = 300 * time.Millisecond
 	maxReconnectInterval     = 60 * time.Second
+
+	// Number of recvonly transceivers pre-allocated on the publisher PC in
+	// single peer connection mode so auto-subscribed media has m-sections
+	// ready in the initial offer.
+	initialMediaSectionsAudio = 3
+	initialMediaSectionsVideo = 3
 )
 
 type RTCEngine struct {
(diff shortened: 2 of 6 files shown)


paulwe (approved):

boks1971 (approved):
A couple of things
1. I had a hack in room.go to enable single peer connection if version is 3.0.0. Did not see it removed in this PR? Maybe I missed it. Can we remove it if not removed?
2. We have `/v0` fallback in JS SDK. If `/v1` signalling fails, SDK does `/v0` fallback to accommodate old OSS deploys. I could not quite follow here if that is done. Can we add it if not?

cnderrauber:
> A couple of things
> 
> 1. I had a hack in room.go to enable single peer connection if version is 3.0.0. Did not see it removed in this PR? Maybe I missed it. Can we remove it if not removed?
> 2. We have `/v0` fallback in JS SDK. If `/v1` signalling fails, SDK does `/v0` fallback to accommodate old OSS deploys. I could not quite follow here if that is done. Can we add it if not?

i thought 1 was intentional so left it there, will remove it if it is useless.

single pc mode has been supported for months ago, does it make sense to update go sdk but keep a outdated server version?

boks1971:
> i thought 1 was intentional so left it there, will remove it if it is useless.

Not sure when we will update version. With this extra flag, that check can be removed for now. When we decide on 3.0 and if that will support only single peer connection, we can enforce that later.

> single pc mode has been supported for months ago, does it make sense to update go sdk but keep a outdated server version?

true, but I think we still get some OSS reports of use of very old servers. Lukas is off this week. We can check with him about it once he is back.

signalling/interfaces.go
```diff
@@ -100,6 +100,10 @@ type ConnectParams struct {
 	// DataEncryptionKeyProvider enables data channel E2EE when set.
 	DataEncryptionKeyProvider e2eetypes.KeyProvider
 
+	// UseSinglePeerConnection enables single peer connection mode in which the
+	// publisher PC is reused for both publishing and receiving media.
```
biglittlebigben, line 104: Should we add a comment on why somebody may or may not want to do that?

PR #908, opened 2026-05-18
TEL-588: expose original raw protocol disconnect reason

hechen-eng:
We currently rely on `OnDisconnectedWithReason` to check whether it's a success/client error/server error. The existing `DisconnectionReason` is not sufficient to cover all reasons. For example,  `ROOM_DELETED`, `SERVER_SHUTDOWN`, `MEDIA_FAILURE` will collapse into `OtherReason`. This PR exposes the original protocol disconnect reason through `DisconnectReason`.

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -138,6 +138,9 @@ const (
 )
 
 // GetDisconnectionReason converts a protocol disconnect reason to a DisconnectionReason.
+// Many protocol reasons (e.g. ROOM_DELETED, SERVER_SHUTDOWN, MEDIA_FAILURE) collapse
+// into OtherReason. New code that needs to distinguish between them should read the
+// raw protocol value via Room.DisconnectReason() instead.
 func GetDisconnectionReason(reason livekit.DisconnectReason) DisconnectionReason {
 	// TODO: SDK should forward the original reason and provide helpers like IsRequestedLeave.
 	r := OtherReason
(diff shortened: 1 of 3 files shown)


anunaym14:
not sure if we should do public api changes

anunaym14:
probably better to just expose the missing/remapped reasons with the existing strucutre. it would be a behavior change but nobody's code will break

hechen-eng:
> probably better to just expose the missing/remapped reasons with the existing strucutre. it would be a behavior change but nobody's code will break

@anunaym14 are you referring to `GetDisconnectionReason`? e.g. add missing mapping between `livekit.DisconnectReason` and `DisconnectionReason`? I am okay with that too, but behavior change may go un-noticed and do harms.

For this API engineHandler::`OnDisconnected`, are we expecting users using it?

anunaym14:
yes I meant that. and I agree that silent failures can be more harmful.

Room basically implements engine handler AFAICT, so it must be getting used

hechen-eng:
> yes I meant that. and I agree that silent failures can be more harmful.
> 
> Room basically implements engine handler AFAICT, so it must be getting used

yeah, it's used by Room, but is it used by anything else? `Room` is also inside server-sdk-go, so it's not a breaking change from that point of view 🤔

anunaym14:
Ah okay, I see what you mean. Sorry, 5am blues.

anunaym14:
I don't really understand why room's OnDisconnect is a public method

callback.go
anunaym14, line 127: Remove the type if it's not used anywhere else?

anunaym14 (approved):
Rest should be good!

callback.go
hechen-eng: I am not sure... it's exported and is used by livekit/sip https://github.com/livekit/sip/blob/main/pkg/sip/room.go#L151. I'll leave it as is for now 👍

PR #911, opened 2026-05-22
sync engine fixes

frostbyte73:
- fixes NTP estimator freezing forever after a sender clock step (now self-heals after 5 consecutive outliers; also floors residStd so float-precision zeros don't disable outlier detection)
- fixes backward RTP timestamps wiping NTP state (was unsigned-wrapping into a 24h delta and tripping the 30s discontinuity reset)
- fixes onSR firing after RemoveTrack
- fixes lastSlewPTS recording the pre-force-correct / pre-monotonicity PTS instead of the actually-emitted PTS

### DIFF
--- a/pkg/synchronizer/syncengine.go
+++ b/pkg/synchronizer/syncengine.go
@@ -211,12 +211,15 @@ func (e *SyncEngine) OnRTCP(packet rtcp.Packet) {
 	// Feed the SR to the session timeline (updates NTP estimator + OWD).
 	e.timeline.OnSenderReport(participantID, trackID, clockRate, sr.NTPTime, sr.RTPTime, now)
 
-	// Call onSR callback if set.
+	// Call onSR callback if set. If the track has been closed concurrently,
+	// drop the callback — the consumer (e.g., tempo controller) may have torn
+	// down by now.
 	st.mu.Lock()
 	onSR := st.onSR
+	closed := st.closed
 	st.mu.Unlock()
 
-	if onSR == nil {
+	if onSR == nil || closed {
 		return
 	}
 
--- a/pkg/synchronizer/syncengine.go
+++ b/pkg/synchronizer/syncengine.go
@@ -93,6 +93,12 @@ type SyncEngine struct {
 
 	startedAt atomic.Int64
 	endedAt   atomic.Int64
+	// ended is the authoritative "End() has been called" signal. It is
+	// distinct from endedAt: endedAt is the PTS-derived end timestamp (zero
+	// if End was called before any track started), whereas ended is set
+	// unconditionally on End() so post-End AddTrack can detect the sealed
+	// state even when the session never produced media.
+	ended atomic.Bool
 
 	// high-water mark for removed tracks, so End() includes their PTS
 	maxRemovedPTS time.Duration
@@ -166,9 +172,12 @@ func (e *SyncEngine) AddTrack(track TrackRemote, participantID string) TrackSync
 		st.startGate = newStartGate(clockRate, track.Kind(), nil)
 	}
 
-	if e.endedAt.Load() != 0 {
+	if e.ended.Load() {
 		// Post-End: the session is sealed. Mark the track closed so every
-		// GetPTS returns io.EOF without emitting media.
+		// GetPTS returns io.EOF without emitting media. Use the dedicated
+		// `ended` flag rather than endedAt: when End() ran with no tracks
+		// initialized yet (startedAt == 0), endedAt stays at 0 since there
+		// is no meaningful PTS-derived timestamp.
 		st.closed = true
 		st.logger.Warnw("AddTrack called after End(); returning closed track", nil)
 	}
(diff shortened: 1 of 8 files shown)


pkg/synchronizer/syncengine_test.go
```diff
@@ -68,8 +68,13 @@ func makeExtPacket(ts uint32, sn uint16, receivedAt time.Time) jitter.ExtPacket
 // --- Tests ---
 
 func TestSyncEngine_ImplementsSyncInterface(t *testing.T) {
-	// Compile-time check that SyncEngine implements Sync.
+	// Compile-time check that SyncEngine implements Sync and that
```
boks1971: should we delete these comments? I think the original comment is not really necessary. But, these are getting very verbose.

pkg/synchronizer/syncengine.go
```diff
@@ -255,33 +319,82 @@ func (e *SyncEngine) OnRTCP(packet rtcp.Packet) {
...
+	e.ended.Store(true)
+
+	// Acquire all per-track locks for a single atomic snapshot + drain pass.
+	// Holding every st.mu through both the max-PTS read AND the drain-ceiling
+	// write closes the TOCTOU in which a track could advance lastPTSAdjusted
+	// between the snapshot loop and the drain-set loop, emitting a packet
+	// past what endedAt promised. GetPTS is the only other st.mu holder, and
+	// it never acquires more than one st.mu at a time, so there is no
+	// deadlock risk in lock-everything order.
 	for _, st := range e.tracks {
```
milos-lk, line 350: I am realizing now that this is iterating map and the order isn't deterministic there. Although this is the only place where we do this now - it's a bit brittle - if we ever need to lock them in other place again - we can easily hit the case of inverse locks acquiring and deadlock. Not sure if this comment will save us - can be easily overlooked. Maybe it's safer to collect them into a slice sorted by e.g SSRC before locking?

pkg/synchronizer/syncenginetrack.go
```diff
@@ -402,5 +531,26 @@ func (st *syncEngineTrack) LastPTSAdjusted() time.Duration {
...
+// Closing only stops FUTURE OnRTCP invocations (which check st.closed before
+// the timeline call). An OnRTCP that has already acquired st.mu and read
+// st.onSR into a local before closeLocked can run will still invoke that
+// captured callback once after close — st.mu serializes the two but the
+// callback function pointer is already copied. Consumers must therefore
+// tolerate one final SR callback per RemoveTrack/Close. Tightening this
+// would require holding st.mu across the user-supplied callback, which is
+// an unacceptable constraint on what the callback may do.
+func (st *syncEngineTrack) closeLocked() {
 	st.closed = true
```
milos-lk, line 562: we are returning EOF from GetPTS when this flag is set. While it's fine for non emitted tracks I am wondering how it would impact this `Drain` call:
https://github.com/livekit/egress/blob/e47b2ac65ed885c216eb9844470516e9b030fd24/pkg/pipeline/source/track_worker.go#L440

since we have remove track just before it...

pkg/synchronizer/syncenginetrack.go
```diff
@@ -209,19 +296,31 @@ func (st *syncEngineTrack) GetPTS(pkt jitter.ExtPacket) (time.Duration, error) {
...
-		st.lastNtpPTS = rawNtpPTS // Always track raw NTP PTS, never corrected
+
+	// Track lastNtpPTS for forward (or first-NTP-packet) updates — using a
+	// backward sample would corrupt the next iteration's expected-jump
+	// baseline. signedRTPDelta >= 0 covers the first call (delta = 0, by way
+	// of initializeLocked setting lastTS to this packet's ts) and forward
+	// packets. hasLastNtpPTS is the canonical "have we recorded a baseline"
+	// flag; lastNtpPTS being zero is also a valid raw NTP value at session
+	// start so we never rely on it as a sentinel.
+	if ntpNowReady && signedRTPDelta >= 0 {
```
milos-lk, line 307: so `lastNtpPTS` is only updated for first/forward packets, but `lastTS`is updated for every emitted packet, including backward/reordered packets - that means after a backward packet, `lastTS` points to the older packet, while `lastNtpPTS` still points to the newer packet before it. In practice jitterbuffer should prevent most backward arrivals, but maybe just to be defensive - guarding the state updates at the bottom of GetPTS could be added so backward packets don't shift the baseline:                                                                                    
```                                                                                                 
  if signedRTPDelta >= 0 {                                                                     
      st.lastTS = ts                                                                           
      st.lastWallPTS = wallPTS                                                                   
      st.lastSlewPTS = preSlewPTS
  }
```
?

pkg/synchronizer/syncengine.go
frostbyte73: I'm not sure that would make any difference unless we end up trying to lock every track somewhere else without holding the SyncEngine lock, which would be broken anyway because you need to hold the lock to read the map

pkg/synchronizer/syncenginetrack.go
frostbyte73: Edit: Claude started replying to these without asking me

This appears unchanged from main, we only close if the track hasn't emitted yet
```
	for _, st := range e.tracks {
		if st.hasEmitted {
			st.maxPTS = maxPTS
			st.maxPTSSet = true
		} else {
			st.closeLocked()
		}
	}
```

milos-lk (approved):

PR #919, opened 2026-06-03
ensure ConnectTimeout is honored when waiting for webrtc connections

davidzhao:


engine.go
```diff
@@ -220,6 +220,11 @@ func (e *RTCEngine) JoinContext(
 	e.token.Store(token)
 	e.connParams = connectParams
 
+	// ConnectTimeout overrides the default joinTimeout.
+	if connectParams != nil && connectParams.ConnectTimeout > 0 {
+		e.joinTimeout = connectParams.ConnectTimeout
```
boks1971, line 225: This could be small, like 3s default. Is that okay?

boks1971 (approved):

cnderrauber (approved):

engine.go
davidzhao: since Go SDK is used in the server environment, I think that's sensible. 15s attempt times is way too long

boks1971: Cross region should still be enough, need to keep an eye

davidzhao: bumped default timeout to 5s

PR #920, opened 2026-06-04
deps: bump livekit/protocol (jwt v5, xsync v4, errors stdlib)

paulwe:
## Summary

Pure dependency bump: `go.mod` / `go.sum` only. The jwt/v5, xsync/v4, and stdlib-errors migrations are entirely contained in `livekit/protocol`; this SDK doesn't reference any of the renamed symbols directly, so no source changes are needed.

Upstream: livekit/protocol#1606


atomic_compat.go
```diff
@@ -0,0 +1,49 @@
...
+
+package lksdk
+
+import (
+	"math"
+	"sync/atomic"
+)
+
+// atomicString is a drop-in replacement for go.uber.org/atomic.String.
+type atomicString struct {
```
biglittlebigben: Why not importing this from protocol/utils?

biglittlebigben (approved):

PR #929, opened 2026-06-24
Reconnect: fail over across regions instead of re-dialing the pinned one

milos-lk:
During a regional outage, egress (which connects through this SDK) failed to reconnect: it kept dialing the region it was originally pinned to until it exhausted its retries, then gave up and ended the recording. Tracing it showed the SDK only ever picked a region on the **initial** connect — the reconnect path reused `e.url` (the region edge URL) and had no notion of regions at all.

This PR teaches the reconnect path to fail over across regions, and cleans up the region machinery along the way so initial connect and reconnect share one source of truth.

A full reconnect no longer re-dials a single URL. It walks an ordered candidate list — the current region first (a single shot, so benign cases like an SFU scale-down stay put), then the server-sorted region list, then the original global URL — and connects to the first that works. If the current region is gone, we step to the next-best one instead of hammering the dead one. A local `tried` set dedupes candidates (the current region is usually also the geo-nearest entry in the list).

Previously the provider both cached regions endpoint *and* tracked which URLs had been handed out, mutating the cached list as it went (`PopBestURL`). Those are two different lifetimes — a shared, long-lived cache vs. transient per-failover progress — and conflating them forced awkward "copy the attempted set forward on refresh" code. The provider is now just a cache: `RegionSettings` returns the immutable list (fetching if stale), and the failover walk owns its own attempt state as a local variable.

When the server asks us to reconnect, it can attach a region list. We now adopt it before reconnecting, so the failover uses the server's authoritative ordering instead of a possibly-stale regions response.

A resume always targets the same node, so on a region outage it would just retry the dead region forever. On a failed resume we now set `requiresFullReconnect`, so the next attempt becomes a full reconnect and region failover can engage.

The reconnect path passed `context.TODO()`, so a blackholed endpoint was only capped by the WebSocket handshake timeout (~45s). With the new multi-candidate walk that would dominate the wall-clock (N candidates × ~45s per sweep). Each candidate is now bounded by `ConnectParams.ConnectTimeout` (falling back to `joinTimeout`), matching what the initial connect already does, so a dead candidate fails fast and we move on.

In addition to that we honor `Cache-Control: max-age` on regions response now.

This mirrors the JS client's reconnect model: 10 sweeps, the same `n² × 300ms` backoff base between sweeps, an explicit per-attempt timeout, and consumption of server-reported regions.

Big thanks to @boks1971 for spotting the issue! 🙏

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -140,9 +140,10 @@ type RTCEngine struct {
 	token      atomic.String
 	connParams *signalling.ConnectParams
 
-	regionProvider *regionURLProvider
-	originalURL    string // the user-provided (global) URL; reconnect fallback
-	cloudHost      string // parsed cloud hostname; empty when not applicable
+	regionProvider          *regionURLProvider
+	originalURL             string      // the user-provided (global) URL; reconnect fallback
+	cloudHost               string      // parsed cloud hostname; empty when not applicable
+	serverDirectedReconnect atomic.Bool // server sent a region list on the last reconnect leave
 
 	joinTimeout time.Duration
 
@@ -939,25 +940,20 @@ func (e *RTCEngine) connectWithRegionFailover() error {
 		return attempt(url)
 	}
 
-	// current region first (benign cases like SFU scale-down stay put), then the
-	// server-sorted region list, then the global URL as a final fallback
-	candidates := []string{e.url}
-	if settings, err := e.regionProvider.RegionSettings(e.cloudHost, e.token.Load()); err != nil {
+	// when the server directed this reconnect with its own region list, honor it
+	// and skip the current region (it told us where to go); otherwise try current
+	// first (benign cases like SFU scale-down stay put)
+	serverDirected := e.serverDirectedReconnect.Swap(false)
+	var settings *livekit.RegionSettings
+	if s, err := e.regionProvider.RegionSettings(e.cloudHost, e.token.Load()); err != nil {
 		e.log.Infow("could not get region settings for reconnect, using current and original url", "error", err)
 	} else {
-		for _, region := range settings.GetRegions() {
-			candidates = append(candidates, region.Url)
-		}
+		settings = s
 	}
-	candidates = append(candidates, e.originalURL)
+	candidates := buildReconnectCandidates(serverDirected, e.url, settings, e.originalURL)
 
-	tried := make(map[string]bool, len(candidates))
 	var lastErr error
 	for _, url := range candidates {
-		if url == "" || tried[url] {
-			continue
-		}
-		tried[url] = true
 		if err := attempt(url); err == nil {
 			return nil
 		} else {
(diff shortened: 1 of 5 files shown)


boks1971 (approved):
lgtm! but would be good to get some more pairs of 👀

biglittlebigben (approved):
LGTM. How hard would it be to add a test for the engine changes?

milos-lk:
> LGTM. How hard would it be to add a test for the engine changes?

asked myself the same question - I think without some refactoring - adding a valuable test that could exercise the flow e2e is going to be tricky - we have real JoinContext which does real dial so without mocking bunch of things not sure how much we can test . One thing that came to my mind is testing this by causing controllable region cut-off on staging and verifying that it recovers - downside of that is that it's hard to automate. But maybe someone has better idea.

cc @lukasIO have we done some testing with similar retry logic inside js sdk?

boks1971:
> > LGTM. How hard would it be to add a test for the engine changes?
> 
> asked myself the same question - I think without some refactoring - adding a valuable test that could exercise the flow e2e is going to be tricky - we have real JoinContext which does real dial so without mocking bunch of things not sure how much we can test . One thing that came to my mind is testing this by causing controllable region cut-off on staging and verifying that it recovers - downside of that is that it's hard to automate. But maybe someone has better idea.
> 
> cc @lukasIO have we done some testing with similar retry logic inside js sdk?

We could potentially add simulation scenario where the server sends `LeaveRequest` dropping self region and forcing the client to go to a different region and exercise this path.

milos-lk:
I like `LeaveRequest`  based simulation scenario - thinking about it - one thing stood out - which I would like to update here - if we do get the list of regions on leave request - we shouldn't do one shot current region retry but proceed with what server provided. Will add that in preparation for the simulation.

engine.go
```diff
@@ -972,6 +968,32 @@ func (e *RTCEngine) connectWithRegionFailover() error {
...
+	if !serverDirected {
+		raw = append(raw, currentURL)
+	}
+	for _, region := range settings.GetRegions() {
+		raw = append(raw, region.Url)
+	}
+	raw = append(raw, originalURL)
+
+	seen := make(map[string]bool, len(raw))
+	out := raw[:0]
```
biglittlebigben: This is probably safe currently, but is it really worth reusing the array here vs using a new one? A future change could easily make this unsafe.

milos-lk: agreed - will update this.

biglittlebigben (approved):

milos-lk:
The existing simulation scenario does what we want as far as I can see `leave_request_full_reconnect`
we probably just need to add trigger here.

davidzhao (approved):
lg. can we add an integration test proving the region fallback? we can mock the initial failure regions

milos-lk:
region walk test set + simulation triggering for other integration tests (e.g egress) added here: https://github.com/livekit/server-sdk-go/pull/932

PR #935, opened 2026-06-27
feat: add WithSettingEngineFunc ConnectOption to customize the SettingEngine

UbiShipDev1:
## Motivation

`NewPCTransport` builds its `webrtc.SettingEngine` internally with no extension point, so callers cannot influence ICE behavior beyond `WithICETransportPolicy` / `WithDisableTURN`.

We run a server-side agent **co-located with the SFU** on a host that has many interfaces (docker bridges, `veth*`, IPv6). pion gathers ICE host candidates from all of them and sometimes selects an unreachable one, silently breaking the subscriber media path (the agent goes "deaf"). The remedy is `SettingEngine.SetInterfaceFilter` / `SetIPFilter`, which the SDK doesn't expose.

## Change

Adds a single, opt-in `ConnectOption`:

```go
// WithSettingEngineFunc customizes the pion SettingEngine before the WebRTC
// API is constructed. Invoked for every PeerConnection (publisher + subscriber).
func WithSettingEngineFunc(fn func(*webrtc.SettingEngine)) ConnectOption
```

`fn` runs after the SDK finishes configuring the `SettingEngine` and before `webrtc.NewAPI`, so callers can layer ICE interface/IP filters, `SetNAT1To1IPs`, etc. without losing the SDK's own settings.

Usage:
```go
room, err := lksdk.ConnectToRoom(url, info, cb,
    lksdk.WithSettingEngineFunc(func(se *webrtc.SettingEngine) {
        se.SetInterfaceFilter(func(name string) bool { return name == "eth0" || name == "lo" })
        se.SetIPFilter(func(ip net.IP) bool { return ip.To4() != nil })
    }),
)
```

## Details
- `signalling.ConnectParams.SettingEngineFunc` carries the hook; `WithSettingEngineFunc` sets it (mirrors the existing option pattern).
- Threaded into both `NewPCTransport` calls (publisher + subscriber) and invoked in `NewPCTransport` just before `webrtc.NewAPI`.
- **No behavior change when unset.** Unit test included (`TestWithSettingEngineFunc`). `go build ./... && go test` green.

Happy to adjust naming or scope to fit your conventions.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -409,6 +409,7 @@ func (e *RTCEngine) createPublisherPCLocked(configuration webrtc.Configuration) 
 		OnRTTUpdate:                e.setRTT,
 		IsSender:                   true,
 		DTLSEllipticCurves:         e.connParams.DTLSEllipticCurves,
+		SettingEngineFunc:          e.connParams.SettingEngineFunc,
 	}); err != nil {
 		return err
 	}
@@ -519,6 +520,7 @@ func (e *RTCEngine) createSubscriberPCLocked(configuration webrtc.Configuration)
 		Interceptors:               e.connParams.Interceptors,
 		IncludeDefaultInterceptors: e.connParams.IncludeDefaultInterceptors,
 		DTLSEllipticCurves:         e.connParams.DTLSEllipticCurves,
+		SettingEngineFunc:          e.connParams.SettingEngineFunc,
 	}); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


cnderrauber (approved):
looks good!

cnderrauber:
@UbiShipDev1 could you please sign our [Contributor License Agreement](https://cla-assistant.io/livekit/server-sdk-go?pullRequest=935)?

PR #936, opened 2026-06-27
feat: auto failover for API with LK Cloud

davidzhao:
automatically fail over 5xx and transport errors

### DIFF
--- a/.github/workflows/test-api.yml
+++ b/.github/workflows/test-api.yml
@@ -14,6 +14,9 @@
 
 name: Test API
 
+permissions:
+  contents: read
+
 on:
   workflow_dispatch:
   push:
--- a/.github/workflows/test-api.yml
+++ b/.github/workflows/test-api.yml
@@ -1,0 +1,57 @@
+# Copyright 2026 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Test API
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  failover:
+    runs-on: ubuntu-latest
+    services:
+      mock-server:
+        image: livekit/test-server:latest
+        ports:
+          - 9999:9999
+          - 10000:10000
+          - 10001:10001
+          - 10002:10002
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
+
+      - name: Set up Opus
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libsoxr-dev libopus-dev libopusfile-dev
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version: 1.26.0
+
+      - name: Wait for mock server
+        run: |
+          for i in $(seq 1 30); do
+            curl -sf http://127.0.0.1:9999/settings/regions >/dev/null && exit 0
+            sleep 1
+          done
+          echo "mock server did not become ready" && exit 1
+
+      - name: Run API tests
+        run: go test -v -count=1 ./apitest/...
--- a/.github/workflows/test-api.yml
+++ b/.github/workflows/test-api.yml
@@ -57,4 +57,4 @@ jobs:
           echo "mock server did not become ready" && exit 1
 
       - name: Run API tests
-        run: go test -v -count=1 ./apitest/...
+        run: go test -v -count=1 -run '^TestAPI_' .
--- a/failover.go
+++ b/failover.go
@@ -51,9 +51,9 @@ const (
 	defaultBackoffBase = 200 * time.Millisecond
 )
 
-// FailoverConfig tunes the region-failover retry loop. A zero value is valid
+// FailoverOptions tunes the region-failover retry loop. A zero value is valid
 // and uses the defaults (auto mode, 3 attempts, 200ms base backoff).
-type FailoverConfig struct {
+type FailoverOptions struct {
 	// Mode selects when failover is active. Defaults to FailoverAuto.
 	Mode FailoverMode
 	// MaxAttempts is the total number of attempts including the first, so the
@@ -64,7 +64,7 @@ type FailoverConfig struct {
 	BackoffBase time.Duration
 }
 
-func (c FailoverConfig) withDefaults() FailoverConfig {
+func (c FailoverOptions) withDefaults() FailoverOptions {
 	if c.MaxAttempts <= 0 {
 		c.MaxAttempts = defaultMaxAttempts
 	}
@@ -76,7 +76,7 @@ func (c FailoverConfig) withDefaults() FailoverConfig {
 	return c
 }
 
-func (c FailoverConfig) enabledFor(hostname string) bool {
+func (c FailoverOptions) enabledFor(hostname string) bool {
 	switch c.Mode {
 	case FailoverOff:
 		return false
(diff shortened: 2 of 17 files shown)


.github/workflows/test-api.yml
```diff
@@ -0,0 +1,57 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  failover:
+    runs-on: ubuntu-latest
+    services:
+      mock-server:
+        image: livekit/test-server:latest
```
davidzhao, line 32: depends on https://github.com/livekit/livekit/pull/4627

failover.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+}
+
+var sharedAPIRegions = newAPIRegionCache()
+
+// apiRegionCache fetches and caches the LiveKit Cloud region list per host for
+// the API failover path. Unlike the signaling regionURLProvider it honors the
+// request scheme (so it works against an http mock server) and forwards the
+// caller's headers (auth plus any custom headers) to the discovery endpoint.
+type apiRegionCache struct {
```
milos-lk: We now have two region caches - apiRegionCache/sharedAPIRegions here and the per-Room regionURLProvider - both fetching /settings/regions, unmarshalling RegionSettings, and parsing Cache-Control via parseRegionSettingsMaxAge. They've already diverged a bit: the provider defaults to a 3s TTL when there's no max-age, while apiRegionCache only caches when max-age > 0, so absent the header it re-fetches on every failover.

Rather than a second implementation, could we extract the shared fetch+parse — something like fetchRegionSettings(scheme, host, headers) (*livekit.RegionSettings, ttl, error) — and have both call it, with the same 3s default? That keeps the lifetimes separate but stops the parsing/TTL logic from drifting apart.

failover.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+func (c FailoverConfig) withDefaults() FailoverConfig {
+	if c.MaxAttempts <= 0 {
+		c.MaxAttempts = defaultMaxAttempts
+	}
+	if c.BackoffBase < 0 {
+		c.BackoffBase = 0
+	} else if c.BackoffBase == 0 {
+		c.BackoffBase = defaultBackoffBase
```
milos-lk: nit: this contradicts with the API doc `Set to 0 to retry without delay` ?

failover.go
```diff
@@ -0,0 +1,352 @@
...
+		if !isRetryable(resp, err) {
+			return resp, err
+		}
+		if attempt == maxAttempts-1 {
+			break // out of attempts; surface the last result
+		}
+
+		// discover regions lazily
+		if !fetchedRegions {
+			settings = t.regions.get(originalScheme, originalHost, req.Header)
```
milos-lk: Should the discovery get/fetch respect the caller's context? Right now RoundTrip calls t.regions.get(originalScheme, originalHost, req.Header) without req.Context(), and fetch uses http.NewRequest on the 5s-timeout client — so it ignores the caller's deadline/cancel.

milos-lk:
just for my education: are all APIs idempotent and safe to be retried on 5xx?

milos-lk:
related to my comment above - just  thinking out loud here
putting failover in the http.RoundTripper is neat and low-touch, but at that layer the failover decision is being made on opaque bytes + an HTTP status. It can read the method name (twirp.MethodName(ctx) rides on the request ctx), but not the typed request or the typed twirp error — so based on what I can see it can't really tell a safe-to-replay read from a mutation, and it ends up retrying every 5xx for every call. Should CreateRoom/StartRoomCompositeEgress/etc. silently replay against another region on a 5xx or dropped connection? That's an at-least-once semantic for operations that were effectively at-most-once before, and the transport can't distinguish them — which makes me think the retry/idempotency policy wants to live somewhere it can see the method and error.
Not sure how all of that reflects observability (if that's a concern now or in future) - would metrics be generic HTTP retries rather than “Egress.StartRoomCompositeEgress failed over from region A to region B?

failover_apitest_test.go
```diff
@@ -0,0 +1,139 @@
...
+	require.NoError(t, err)
+	require.Equal(t, "api-test", room.Name, "the mock echoes the request name")
+	require.NotEmpty(t, room.Sid)
+}
+
+func TestAPI_PrimaryUnavailable(t *testing.T) {
+	client := lksdk.NewRoomServiceClient(testServerURL(t), "devkey", "secret")
+	ctx := failoverCtx(t, lksdk.FailoverOn, map[string]string{hdrFailRegions: "0"})
+	_, err := client.CreateRoom(ctx, &livekit.CreateRoomRequest{Name: "api-test"})
+	require.NoError(t, err, "should fail over to a healthy region")
```
boks1971, line 90: can this test the exact region that it connects to? Guess not as `CreateRoomResponse` does not have a region indication.

failover.go
```diff
@@ -0,0 +1,360 @@
...
+// FailoverOptions tunes the region-failover retry loop. A zero value is valid
+// and uses the defaults (auto mode, 3 attempts, 200ms base backoff).
+type FailoverOptions struct {
+	// Mode selects when failover is active. Defaults to FailoverAuto.
+	Mode FailoverMode
+	// MaxAttempts is the total number of attempts including the first, so the
+	// original host plus (MaxAttempts-1) fallback regions. Defaults to 3.
+	MaxAttempts int
+	// BackoffBase is the initial delay before the first retry; each subsequent
+	// retry doubles it. Defaults to 200ms. Set to 0 to retry without delay.
```
boks1971: Probably not allow this no delay if this is at API level. Have horror stories at a previous job where a no back off retry kept piling up and eventually killing the service.

boks1971: I guess users can still bypass and do their own back off, but at least not provide that option here?

failover.go
```diff
@@ -0,0 +1,360 @@
...
+		status := 0
+		if resp != nil {
+			status = resp.StatusCode
+		}
+		logger.Warnw("livekit API request failed, retrying with fallback url", err,
+			"failedUrl", scheme+"://"+host, "fallbackUrl", nextScheme+"://"+nextHost,
+			"attempt", attempt+1, "maxAttempts", maxAttempts, "status", status)
+
+		drainResponse(resp)
+		if !sleepCtx(req.Context(), cfg.BackoffBase<<uint(attempt)) {
```
boks1971: should there some upper bound to protect against badly configured back off?

boks1971 (approved):

failover.go
davidzhao: yeah, realizing you just added something similar, I'll refactor them

davidzhao:
> related to my comment above - just thinking out loud here putting failover in the http.RoundTripper is neat and low-touch, but at that layer the failover decision is being made on opaque bytes + an HTTP status. It can read the method name (twirp.MethodName(ctx) rides on the request ctx), but not the typed request or the typed twirp error — so based on what I can see it can't really tell a safe-to-replay read from a mutation, and it ends up retrying every 5xx for every call. Should CreateRoom/StartRoomCompositeEgress/etc. silently replay against another region on a 5xx or dropped connection? That's an at-least-once semantic for operations that were effectively at-most-once before, and the transport can't distinguish them — which makes me think the retry/idempotency policy wants to live somewhere it can see the method and error. Not sure how all of that reflects observability (if that's a concern now or in future) - would metrics be generic HTTP retries rather than “Egress.StartRoomCompositeEgress failed over from region A to region B?

In this case, all requests should be retried. our server should be as idempotent as possible. that would be the right design patterns here. otherwise, if a request returns 500, how should we react to it? the contract says the API errored out and the most natural thing for any end user to handle is to retry it.

for observability, we should be able to see the HTTP path to determine the method/etc.

failover.go
davidzhao: it shouldn't.. this is an internal request to discover regions, we'll limit it to a much shorter timeout. so for example, if someone was doing a CreateSIPParticipant with a 30s timeout, we do not want region refresh to take the same 30s.

failover.go
davidzhao: great point.. I've removed these options, since this is internal, I've kept it hard coded. we can always introduce new options later

davidzhao:
@milos-lk updated the implementation to address feedback. now we have a consolidated region cache.

failover.go
milos-lk: Totally agree the discovery fetch shouldn't ride the caller's 30s — capping it short is right. One clarification though: threading the context doesn't undo that cap. http.NewRequestWithContext(ctx, …) on the Timeout: 2s client is bounded by min(2s, remaining deadline), not the max  - the client derives its deadline from the request ctx, so whichever fires first wins - right? Your CreateSIPParticipant/30s case still gets capped at 2s; the 30s never applies.
The reason to still pass ctx is the two cases the fixed 2s alone doesn't cover:
 - Cancellation - if the caller cancels (client disconnected, op aborted, process draining), the in-flight discovery keeps burning up to 2s on an abandoned request instead of stopping right away.
 - Sub-2s deadlines (maybe that's not too frequent in practice - but theoretical possibility) — a caller with a 500ms timeout: the primary fails at 500ms, then discovery still runs up to 2s, so the call returns ~2.5s and the result gets discarded anyway.

failover.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	return resp != nil && resp.StatusCode >= 500
+}
+
+// nextRegion returns the first region whose host has not yet been attempted.
+func nextRegion(settings *livekit.RegionSettings, attempted map[string]struct{}) (scheme, host string, ok bool) {
+	if settings == nil {
+		return "", "", false
+	}
+	for _, region := range settings.Regions {
```
biglittlebigben, line 334: Is the Regions slice returned in settings in the same order across projects? If so, should we shuffle it?

biglittlebigben:
> just for my education: are all APIs idempotent and safe to be retried on 5xx?

We tell our customers to retry on 5xx without any extra caveat. We should solve any bug that make 5xx retries no idempotent if we find any. Timeouts particularly could be problematic (ie StartEgress?)

biglittlebigben (approved):

failover.go
davidzhao: it's returned depending on distance to the user, and not shuffled. since this is also optimizing for connection times

failover.go
davidzhao: great points on both, I've added the following:
- if request timeout is under 5s, then we do not retry (too short of timeout)
- figure out a timeout per request, instead of inheriting the context for subsequent retries

PR #937, opened 2026-06-29
feat: OnRoomMovedWithSID

anunaym14:


callback.go
```diff
@@ -172,6 +172,7 @@ type RoomCallback struct {
 	OnRoomMetadataChanged     func(metadata string)
 	OnRecordingStatusChanged  func(isRecording bool)
 	OnRoomMoved               func(roomName string, token string)
+	OnRoomMovedWithSID        func(roomName string, roomSID string, token string)
```
anunaym14, line 175: Maybe it should be Sid and not SID matching the Room struct

boks1971 (approved):

callback.go
biglittlebigben: Proto generated code will always use strict camel case. We tend to keep appreciation all caps in our own naming.

PR #938, opened 2026-06-29
fix: update room name/sid on move

anunaym14:


room.go
```diff
@@ -1316,6 +1316,12 @@ func (r *Room) OnRoomMoved(moved *livekit.RoomMovedResponse) {
 	r.log.Infow("room moved", "newRoom", moved.Room.Name)
 	r.OnRoomUpdate(moved.Room)
```
anunaym14, line 1316: this method does call setSid: https://github.com/livekit/server-sdk-go/blob/88d5310417326f3599b82592d0232ced8879f5d8/room.go#L1306

but it doesn't update anything except for the first time: https://github.com/livekit/server-sdk-go/blob/88d5310417326f3599b82592d0232ced8879f5d8/room.go#L849

davidzhao (approved):
lg, can we add a test to guard for regression?

room.go
```diff
@@ -1303,7 +1303,6 @@ func (r *Room) OnRoomUpdate(room *livekit.Room) {
 		r.activeRecording = room.ActiveRecording
 	}
 	r.lock.Unlock()
-	r.setSid(room.Sid, false)
```
milos-lk, line 1306: Possibly off-base here, so push back if I'm misreading it.

The OnRoomMoved override makes sense to me. What I'm less sure about is this removal. That method also runs for the plain RoomUpdate signal and on reconnect, not just moves, and I think (not certain) that the first time it fires might be how a late-assigned SID gets delivered, given SID() blocks on sidReady and the doc mentions the SID can take a couple seconds after join. If that's the case, dropping it could leave sidReady unclosed for rooms that join without a SID and SID() callers block forever?

But I haven't confirmed whether the JoinResponse always carries a non-empty SID, if it does, this is a non-issue and ignore me.

boks1971: You are correct @milos-lk . SID can come late. I think this cannot be removed.

anunaym14: https://github.com/livekit/server-sdk-go/pull/939

Also added a test for this

PR #942, opened 2026-07-02
Refactor connection state machine.

boks1971:
A rather high diff PR. But, the recommended sequence
- connection_manager.go -> Splits out the connection states o Generates a plan for connection sequence for each state
- room.go -> removed the initial connection loop. This has moved into `join()` in engine.go. That function is used for full reconnect also.
- engine.go -> clean up connection stuff and moved to connection_manager.go. Each stage asks connection manager for a plan and gets ordered list of servers to try with backoffs, validation timeout etc. Ensure context is threaded through and proper deadlines are set.

Please note that there is some repetition of code
- in building plans for different state in connection_manager.go
- between `join()` and `resumeConnection()` when executing the plan in engine.go But, kept them for easier readability. Can consolidate it later once it is stable.

Once again requested Claude for tests and it wrote some exhaustive tests.

@milos-lk I probably removed/reworked some stuff you did just last week. Sorry. Felt that the whole thing was getting entangled and this is an attempt to hopefully make it cleaner.

For manual testing, published video from livekit-cli and had code in there to simulate migration (Go SDK does not support migration, so this also becomes a full reconnect) or node failure or leave request full reconnect every 10 seconds and ensured that it reconnected and published (found an issue with dynacast while doing this which I fixed in https://github.com/livekit/server-sdk-go/pull/941)


milos-lk:
no worries - my change from the last week allowed us to ship that fixed region failover logic in some of services. I am going to take a look to check what this buys us.

milos-lk:
Since ConnectionState() is now derived from the connection manager, what moves the state machine to Disconnected on a client-initiated leave? setDisconnected() seems to only be called from OnLeave on a server LeaveRequest_DISCONNECT, but room.Disconnect() → cleanup() → engine.Close() never touches the connection manager, so after a local disconnect, ConnectionState() would keep returning Connected? Same question for the give-up path: when the reconnect loop exhausts retries and fires OnDisconnected, the state stays Reconnecting? Would it make sense to call setDisconnected() from engine.Close()?

boks1971:
> Since ConnectionState() is now derived from the connection manager, what moves the state machine to Disconnected on a client-initiated leave? setDisconnected() seems to only be called from OnLeave on a server LeaveRequest_DISCONNECT, but room.Disconnect() → cleanup() → engine.Close() never touches the connection manager, so after a local disconnect, ConnectionState() would keep returning Connected? Same question for the give-up path: when the reconnect loop exhausts retries and fires OnDisconnected, the state stays Reconnecting? Would it make sense to call setDisconnected() from engine.Close()?

Good catch! I did that last as I saw there was another connection state and tried to re-use. Will fix the disconnected bit.

connection_manager.go
```diff
@@ -0,0 +1,414 @@
...
+		return false
+	}
+
+	// if not connected, cannot resume, so no-op
+	if c.state != connectionManagerStateConnected {
+		return false
+	}
+
+	// if already resuming, do not take settings that are nil as some internal paths might do a resume without regions
+	if c.state == connectionManagerStateResuming {
```
milos-lk, line 245: dead code? resuming is != connectionManagerStateConnected - so the check above returns first?

boks1971: Good catch, I need to move that up before the previous check. There are cases where the resume happens multiple times. It can happen internally and it can also come from server. So, want to latch the server given list if available.

connection_manager.go
```diff
@@ -0,0 +1,414 @@
...
+	return true
+}
+
+func (c *connectionManager) setReconnecting(regionSettigs *livekit.RegionSettings) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.regionSettings = utils.CloneProto(regionSettigs)
+	c.updateState(connectionManagerStateReconnecting)
+	return true
```
milos-lk: What prevents a leave: RECONNECT arriving during the initial join (state still Initial - the read loop starts in OnJoinResponse, before setConnected) from spawning the reconnect goroutine here and racing the in-flight join()? The old code gated handleDisconnect on hasConnected. setResuming has a state guard but setReconnecting returns true unconditionally.

boks1971: There could be a small window there. Will look at plugging that.

engine.go
```diff
@@ -793,55 +836,82 @@ func (e *RTCEngine) readDataPacket(msg webrtc.DataChannelMessage) (*livekit.Data
...
+		if !e.connectionManager.setReconnecting(regionSettings) {
+			return
 		}
+	} else {
+		if !e.connectionManager.setResuming(regionSettings) {
+			return
+		}
+	}
+
+	if !e.reconnecting.CompareAndSwap(false, true) {
```
milos-lk: If the goroutine is between setResumed (state=Connected) and its defer e.reconnecting.Store(false) when a new transport-close arrives, setResuming succeeds, this CAS fails, and the goroutine exits, leaving state=Resuming with no worker. What re-arms recovery at that point? setResuming will refuse every subsequent resume-type event (state != Connected), and a leave: RECONNECT needs a signal connection we no longer have.

connection_manager.go
```diff
@@ -0,0 +1,414 @@
...
+	c.regionSettings = utils.CloneProto(regionSettigs)
+	c.updateState(connectionManagerStateResuming)
+	return true
+}
+
+func (c *connectionManager) setReconnecting(regionSettigs *livekit.RegionSettings) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.regionSettings = utils.CloneProto(regionSettigs)
```
milos-lk: maybe having something like:
```
    if regionSettigs != nil {
        c.regionSettings = utils.CloneProto(regionSettigs)
    }
```
would help with rare cases where leave reconnects lands during the failing resume - we won't overwrite server directed ordering?

connection_manager.go
```diff
@@ -0,0 +1,414 @@
...
+		"old", c.state,
+		"new", state,
+		"regionSettings", protoLogger.Proto(c.regionSettings),
+		"currentRegion", protoLogger.Proto(c.connectedRegion),
+	)
+	c.state = state
+}
+
+func (c *connectionManager) getConnectionPlan() ([]connectionAttemptParams, error) {
+	c.mu.RLock()
```
milos-lk, line 397: holding the lock while performing HTTP requests (I think the timeout there is 5s) is making me somehow uncomfortable.
Should we snapshot the state and unlock early here?

while holding the lock for extended period of time - writers might start piling up - e.g setResuming/Reconnecting from handleDisconnect(), setToken() - that could then cause all subsequent RLock to block - currentState(), getConnectParams()..
maybe could even escape to the room API - haven't followed all possibilities there.

connection_manager.go
```diff
@@ -0,0 +1,414 @@
...
+		return false
+	}
+
+	c.regionSettings = nil
+	c.connectedRegion = utils.CloneProto(region)
+	c.updateState(connectionManagerStateConnected)
+	return true
+}
+
+func (c *connectionManager) setResuming(regionSettigs *livekit.RegionSettings) bool {
```
milos-lk: nit: typo regionSettigs

engine.go
```diff
@@ -851,46 +921,108 @@ func (e *RTCEngine) handleDisconnect(fullReconnect bool) {
...
-	}
 
-	if err = e.waitUntilConnected(); err != nil {
-		return err
+		// restore steady state so subsequent resumes start from Connected (and pick
+		// up fresh server region settings) and connectedRegion tracks where we are.
+		// setResumed is a no-op if a reconnect was requested while resuming, so it
+		// won't clobber a pending full reconnect.
+		e.connectionManager.setResumed(attempt.region)
+		e.engineHandler.OnResumed()
```
milos-lk, line 1016: should this be:
```
if e.connectionManager.setResumed(attempt.region) {
    e.engineHandler.OnResumed()
}
```
I can't put the finger yet - but it feels to me that if left unguarded we might have `OnReconnecting -> OnReconnected -> OnReconnecting ` sequence.

milos-lk:
Was dropping the current region from the reconnect plan intentional? The resuming plan falls back to connectedRegion when there's no server list, but the reconnecting plan goes straight to discovery + original, so a benign reconnect (SFU scale-down, no regions in the leave) leads with the geo-nearest discovery entry rather than the region we're in. For a client that previously failed over, that means migrating back (or burning a connect timeout on a still-dead region), if the discovery fetch fails, the plan is just the original URL.

connection_manager.go
boks1971: sorry @milos-lk I am not able to understand the concern here. Can you elaborate on the sequence you are thinking about?

This is called when leave reconnect is received. So, we want to take the regions from it. If it is `nil`, this will get cleared, but it is okay. We can get it from region URL provider.

connection_manager.go
boks1971: Good catch. I missed the fact that region URL provider can do a HTTP GET. Will re-think this. Snapshotting should be fine I think.

engine.go
boks1971: oh boy, no race condition gets by you, does it? 🙂 

Thank you for the catch. This was one piece I struggled with a bunch and trimmed down as much as I can. Will check how to plug the hole.

engine.go
boks1971: I think it is okay. It will be `OnResuming` -> `OnResumed` -> `OnReconnecting` -> `OnReconnected` if the full reconnect immediately follows a resume.

boks1971:
> Was dropping the current region from the reconnect plan intentional? The resuming plan falls back to connectedRegion when there's no server list, but the reconnecting plan goes straight to discovery + original, so a benign reconnect (SFU scale-down, no regions in the leave) leads with the geo-nearest discovery entry rather than the region we're in. For a client that previously failed over, that means migrating back (or burning a connect timeout on a still-dead region), if the discovery fetch fails, the plan is just the original URL.

No solid reason one way or the other. If client had gone away to a different region, coming back to a closer region might be better assuming it is not dead. Can put that back before discovery list like the resume plan.

connection_manager.go
boks1971: Addressed in https://github.com/livekit/server-sdk-go/pull/942/changes/cf41a2741e90102fd011c504a7942f3746fe5cee. Will appreciate if you can have a look.

engine.go
milos-lk: yeah - state machines are always tricky - I like what you did here - turning bunch of implicit flags into the explicit state machine - makes it easier to spot these 👍

connection_manager.go
milos-lk: maybe it's fine - I was thinking about the following - If a leave-reconnect with regions arrives while a resume is failing, setReconnecting(regions) stores them (CAS fails, returns), then the resume goroutine's escalation e.connectionManager.setReconnecting(nil) overwrites them with nil.. probably edge case and not that important.

boks1971: aha, got it, good call, will look at it.

connection_manager.go
milos-lk: perfect thanks - that fixes it as far as I can see and also covers deadcode and current region first 👍 
one minor nit (could be preference / taste related - so feel free to ignore it): now that the planners take everything as parameters,  buildConnectionPlan uses no receiver state at all, and the three planners only use c.regionProvider/c.log. Passing those in (or bundling the snapshot into a small struct) would make "planners never touch mutable manager state" enforced by the compiler rather than by convention.

boks1971: Good idea, done in https://github.com/livekit/server-sdk-go/pull/942/changes/94377099fafcb3aed0dcae4412187af076899375

milos-lk (approved):

connection_manager.go
boks1971: Addressed in https://github.com/livekit/server-sdk-go/pull/942/changes/9e3378c3be98ad554e5eb31787840376ac8b6698

connection_manager.go
```diff
@@ -0,0 +1,472 @@
...
+// setResumed restores the Connected state after a successful resume so the next
+// resume starts fresh. It is a no-op unless still Resuming: if a reconnect was
+// requested while the resume was in progress, the state is left Reconnecting so
+// the pending full reconnect proceeds rather than being clobbered.
+func (c *connectionManager) setResumed(region *livekit.RegionInfo) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Disconnected is terminal; never transition out of it
+	if c.state == connectionManagerStateDisconnected {
```
biglittlebigben: Is this check needed considering the check right below?

connection_manager.go
```diff
@@ -0,0 +1,472 @@
...
+				params.log.Infow(
+					"first region in settings does not match connected region for resume",
+					"firstRegion", protoLogger.Proto(regions[0]),
+					"connectedRegion", protoLogger.Proto(params.connectedRegion),
+				)
+			}
+		}
+
+		// server sent list via LeaveRequest, try those
+		regionsToTry = append(regionsToTry, params.regionSettings.GetRegions()...)
```
biglittlebigben, line 482: I didn't realize resuming worked across region. Are the participant state and publications kept in this case?

biglittlebigben (commented):
I only looked at part of the patch and had a couple of questions. Don't wait for me to merge this.

boks1971:
> I only looked at part of the patch and had a couple of questions. Don't wait for me to merge this.

Thank you @biglittlebigben . This is still WIP as I am still working on a couple of holes identified by @milos-lk . So, please feel free to send more feedback if you get a chance to look more. But, if I get to a point of things looking fine, I will merge and can still accept feedback and fix in a subsequent PR.

connection_manager.go
boks1971: Yeah, not technically needed, mostly for consistency.

connection_manager.go
boks1971: Yeah, it works via migrator service + director.

The whole migration sequence is a delicate dance between three parties
- client
- old node
- new node

client sends its state, new node asks for state from old node, reconciles it and migrates participant to new node. If old node is gone, it is not possible irrespective of new node's location (same region or different region.

connection_manager.go
boks1971: Addressed this in https://github.com/livekit/server-sdk-go/pull/942/changes/db49846e3d08e6ff422386b383f4a0f64ea21a02 @milos-lk . Also included an idea in there. We can potentially consider that later. Let me know what you think.

engine.go
boks1971: Went to the Claude on this @milos-lk . As usual, it wrote a lot comments. I tried to tone it down, but it might still be verbose. Let me know your thoughts - https://github.com/livekit/server-sdk-go/pull/942/changes/52c442f36b59bfc71d2083b94e9f06b2e3d316a4

boks1971:
@milos-lk @biglittlebigben Going to merge this. But, please let me know if you have further feedback and I will catch it in a subsequent PR.

PR #943, opened 2026-07-03
correctly handle whatsapp timeout

davidzhao:


paulwe (approved):

failover_internal_test.go
```diff
@@ -52,6 +53,83 @@ func TestFailoverAttempts(t *testing.T) {
...
+		defer cancel()
+		assertBudget(t, ctx, defaultRingingTimeout+ringingTimeoutMargin)
+	})
+	t.Run("no deadline raised above ringing", func(t *testing.T) {
+		ctx, cancel := dialContext(context.Background(), ring(40*time.Second))
+		defer cancel()
+		assertBudget(t, ctx, 40*time.Second+ringingTimeoutMargin)
+	})
+
+	// A caller deadline shorter than the dial budget is extended (the fix): a
```
boks1971: I have been asking Claude to make comments without reference to old code or existing bugs or any reference and keep it functional/factual. This is going to be confusing for people who read this three months from now (actually even now, I don't know what the `fix` here means :-) )

boks1971 (approved):

failover_internal_test.go
davidzhao: great point.. will clean it up

PR #947, opened 2026-07-07
Record subscription intent and reconcile until converged

milos-lk:
UpdateSubscription signal messages are fire-and-forget: SetSubscribed sends one message and keeps no record of it. If that message is lost (stalled controller, dropped relay stream, reconnect race), desired and actual subscription state diverge permanently - the SDK waits forever and sync state on resume only replays subscriptions that already attached.

- record desired subscription state per track in SetSubscribed
- reconcile loop per room re-sends UpdateSubscription (batched, with backoff) until the subscription attaches or is cleared; started lazily on the first recorded desire so rooms that never subscribe selectively never spawn it, stopped on room cleanup
- unsubscribes have no client-side convergence observable (the attached track is never cleared), so they are re-asserted a bounded number of times per connection epoch instead of reconciled to convergence
- sendSyncState replays recorded intent, not just attached tracks
- reconcile kicked (and attempt counters reset) after resume and full reconnect
- handle SubscriptionResponse errors: stop re-asserting rejected subscriptions and fire OnTrackSubscriptionFailed instead of timing out; routed via a new optional SubscriptionResponseProcessor interface so existing SignalProcessor implementations keep compiling

No breaking API changes

### DIFF
--- a/subscriptiondesire.go
+++ b/subscriptiondesire.go
@@ -1,0 +1,273 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lksdk
+
+import (
+	"sync"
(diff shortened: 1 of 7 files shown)


subscriptiondesire.go
```diff
@@ -0,0 +1,273 @@
...
+			due, sub, attempts := store.dueAttempt(sid, pub.IsSubscribed(), force)
+			if !due {
+				continue
+			}
+			if sub {
+				subscribe[pub.participantID] = append(subscribe[pub.participantID], sid)
+			} else {
+				unsubscribe[pub.participantID] = append(unsubscribe[pub.participantID], sid)
+			}
+			r.log.Infow(
```
boks1971, line 232: maybe log this outside the loops?

boks1971 (approved):

PR #949, opened 2026-07-08
version bump to 2.18, readme updates

davidzhao:


paulwe (approved):

README.md
```diff
@@ -50,57 +50,130 @@ func getJoinToken(apiKey, apiSecret, room, identity string) (string, error) {
...
 ```
 
-## RoomService API
+## Authentication
 
-RoomService gives you complete control over rooms and participants within them. It includes selective track subscriptions as well as moderation capabilities.
+Every request to the server APIs is authenticated. `LiveKitAPI` (and each service client) supports two modes:
+
+- **API key & secret** — recommended for backend use. The SDK signs a short-lived token per request from your key and secret. Keep your API secret on the server; never ship it to a client.
+- **Access token** — for frontend / client-side use, where the API secret must not be exposed. Pass a pre-signed [access token](https://docs.livekit.io/frontends/reference/tokens-grants/) that already carries the grants for the operations you'll perform; the SDK sends it verbatim. Mint it on your backend and hand it to the client.
```
boks1971, line 58: Do we need something for generating a token with right permissions for different APIs if using this option?

boks1971 (approved):

README.md
davidzhao: since this is an advanced use case, I'm leaving it alone for now.. just making it so it's possible to do so.

the main use case is on JS and other front-end clients, where the server SDK is now usable within the client SDK, and be able to inherit its credentials

PR #951, opened 2026-07-15
chore(cloudagents): inject attributes header in byoc deploys

rektdeckard:
https://linear.app/livekit/issue/AP-956/byoc-deploy-attributes-are-not-persisted-on-agent-versions

What: injects any user-defined deployment attributes as a request header (`X-LIVEKIT-AGENT-ATTRIBUTES`) after completing image upload in a BYOC deployment.

BYOC agent deployments are currently a 2-step process:
1. SDK requests a push endpoint from cloud-agents service (generated by `crane`) to upload an image to
2. SDK calls cloud-agents`/v2` callback endpoint when the push is complete

The cloud-agents service then creates a record of the agent version and spins it up. In order to support custom user-defined attributes in BYOC cloud agent deployments, we have to forward them to cloud-agents svc *after* pushing the image, because this transaction is stateless from the perspective of cloud-agents svc.

In the Non-BYOC case, which is very similar, we had been appending these attributes as URL parameters. For consistency's sake, I have now standardized both cases to use the header value (with backward compatibility for the parameter).

REF: https://github.com/livekit/cloud-agents/pull/736, https://github.com/livekit/livekit-cli/pull/914

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -140,7 +140,7 @@ require (
 	golang.org/x/tools v0.45.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260526163538-3dc84a4a5aaa // indirect
-	google.golang.org/grpc v1.81.1 // indirect
+	google.golang.org/grpc v1.81.1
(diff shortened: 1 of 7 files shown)


yoonsio:
Can we do `X-LIVEKIT-AGENT-VERSION-ATTRIBUTES` instead? We are going to have support for other attributes in the future.

yoonsio (approved):

rektdeckard:
@yoonsio sure, makes sense

PR #960, opened 2026-08-10
Fix deadlock between Synchronizer.End() and sender report processing

milos-lk:
We found long-lived egress handlers wedged at EOS with goroutine dumps showing a classic ABBA lock cycle in pkg/synchronizer: `Synchronizer.End()` holds the synchronizer write lock and calls `participantSynchronizer.getMaxPTSAdjusted()`, which needs the participant mutex — while an RTCP worker inside `participantSynchronizer.onSenderReport` holds the participant mutex and calls `onSenderReportWithRebase → getSynchronizerDeadline → getExternalMediaDeadline`, which needs the synchronizer lock. Neither can proceed, and every other synchronizer user (`GetEndedAt`, `GetPTS`, other RTCP workers) piles up behind them. The window is one in-flight sender report at the instant `End()` fires, and it requires the rebased sender report mode plus a configured `MediaRunningTime`.

This is the same deadlock class #841 fixed at the track-lock level, one lock-level up: #865 added the deadline fetch to the sender report path and correctly placed it before the track lock per #841's rule, but the call site can't see that its caller holds the participant lock. Since call-site discipline has now failed twice, this PR fixes the specific inversion and removes the discipline requirement entirely.

`participantSynchronizer.onSenderReport` now looks up the track under the participant lock, releases it, and then processes the report — the same pattern `OnRTCP` already uses at the synchronizer level. And `getExternalMediaDeadline` is now lock-free: the `MediaRunningTime` callback lives in an `atomic.Pointer` (kept in sync by the option and by `SetMediaRunningTime`), and the lazily-set external media start time is an `atomic.Pointer[time.Time]` set once via CAS, preserving the monotonic clock reading so wall-clock adjustments can't move the deadline. With no code path acquiring the synchronizer lock from under a child lock, `End()`'s parent→child nesting is acyclic by construction and stays unchanged.

The new regression test spins up sender report processing concurrently with `End()` across many fresh synchronizers; it reliably deadlocks on main within its watchdog and passes with this change. The full package passes `go test -race`.

### DIFF
--- a/pkg/synchronizer/synchronizer_test.go
+++ b/pkg/synchronizer/synchronizer_test.go
@@ -333,10 +333,6 @@ func TestEnd_RemovedTracks_StillContributeDuration(t *testing.T) {
 		"duration must reflect removed track's PTS; got %v", duration)
 }
 
-// Regression: End() holds the synchronizer lock while calling into participant
-// synchronizers, while sender report processing holds the participant lock and
-// reads the external media deadline from the synchronizer — an ABBA deadlock if
-// either side nests the other's lock.
 func TestEnd_NoDeadlockWithConcurrentSenderReports(t *testing.T) {
 	done := make(chan struct{})
 	go func() {
--- a/pkg/synchronizer/synchronizer_test.go
+++ b/pkg/synchronizer/synchronizer_test.go
@@ -1,6 +1,7 @@
 package synchronizer_test
 
 import (
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/synchronizer/synchronizer_test.go
```diff
@@ -331,3 +332,60 @@ func TestEnd_RemovedTracks_StillContributeDuration(t *testing.T) {
 	require.Greater(t, duration, time.Second,
 		"duration must reflect removed track's PTS; got %v", duration)
 }
+
+// Regression: End() holds the synchronizer lock while calling into participant
```
boks1971: maybe remove the word `Regression`. Won't mean much out of the context of this PR.

boks1971: Actually, the whole comment is referring to previous behaviour. Won't make sense for new readers. Maybe make the whole thing a PR comment and not a code comment.

boks1971 (approved):

pkg/synchronizer/synchronizer_test.go
milos-lk: true - getting rid of that

milos-lk: done

PR #963, opened 2026-08-14
separate queue waiting messages from build logs

andrewnitu:
## Goal
we are introducing a queue system to agent builds for periods of bursty load. The server will now send status updates  regarding the queue to the client. However, it does not make sense to include these in the build logs directly.

Thus, introduce a header `X-LIVEKIT-BUILD-PROTOCOL` that indicates to the server the client is capable of receiving queue messages separately. The queue messages are rendered separately from the build logs in the client.

### DIFF
--- a/pkg/cloudagents/build.go
+++ b/pkg/cloudagents/build.go
@@ -16,6 +16,7 @@ package cloudagents
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


biglittlebigben:
Can you please provide a context on what this PR tries to achieve in the PR description?

yoonsio (approved):
LGTM! Do we have any backward compatibility concerns?

PR #978, opened 2026-08-25
fix: guard room ConnectionState nil engine

mariusgassen:
* with the refactoring of reading the connection state of a room directly from its non-exposed `engine`, that will panic if `nil` when the connection state is read
* Current workaround: Check for an exposed `LocalParticipant`

### DIFF
--- a/room.go
+++ b/room.go
@@ -536,6 +536,10 @@ func (r *Room) ConnectionState() ConnectionState {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	if r.engine == nil {
+		return ConnectionStateDisconnected
+	}
+
 	switch r.engine.currentState() {
 	case connectionManagerStateInitial, connectionManagerStateClosed:
 		return ConnectionStateDisconnected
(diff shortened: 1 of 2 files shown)


room.go
```diff
@@ -536,6 +536,10 @@ func (r *Room) ConnectionState() ConnectionState {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	if r.engine == nil {
```
cnderrauber, line 539: `r.engine` is initiated in `NewRoom` and has never been cleared/reassign, it should not be nil. Could you share the panic stack?

mariusgassen: We wrapped the room into a custom struct with business logic on it.

In a test we instantiate it with a `&lksdk.Room{}` instead of using `NewRoom` - this will cause the panic because the engine is never initialized.

I fixed our tests by using `NewRoom(nil)`, however I thought a nil guard wouldn't hurt.

cnderrauber:
It is not a problem in this SDK.

PR #987, opened 2026-09-02
Add data track data channel

ladvoc:
Adds the `_data_track` data channel alongside the existing lossy and reliable channels, forwards incoming packets to the room, and adds a sender that paces frames onto the channel while keeping only the freshest one. This is not integrated with anything yet.

Closes BOT-541


datatracksender.go
```diff
@@ -0,0 +1,125 @@
...
+		done:   make(chan struct{}),
+	}
+}
+
+func (s *dataTrackSender) setLogger(log protoLogger.Logger) {
+	s.log = log
+}
+
+func (s *dataTrackSender) send(frame dataTrackFramePackets) {
+	s.once.Do(func() { go s.run() })
```
boks1971, line 56: probably make a `start()` call? It does only once, but has to check on every send if this is already done.

datatracksender.go
```diff
@@ -0,0 +1,125 @@
...
+	if dropped := s.push(frame); dropped != nil {
+		s.log.Debugw("dropping data track frame", "packets", len(dropped))
```
boks1971, line 58: :nit: `numPackets` key

datatracksender.go
boks1971: guess it can be started from the constructor too. Makes things simple.

engine.go
```diff
@@ -557,6 +566,19 @@ func (e *RTCEngine) createPublisherPCLocked(configuration webrtc.Configuration)
...
+		MaxRetransmits: new(uint16),
+	})
+	if err != nil {
+		e.dclock.Unlock()
+		return err
+	}
+	e.dataTrackDC.OnMessage(e.handleDataTrackPacket)
+	e.dataTrackDC.SetBufferedAmountLowThreshold(dataTrackBufferedAmountLowThreshold)
+	e.dataTrackDC.OnBufferedAmountLow(e.dataTrackSender.wake)
+	e.dataTrackDC.OnOpen(e.dataTrackSender.wake)
```
boks1971, line 581: I made a note about starting the data track sender, this is another place it can be started I think.

datatracksender.go
```diff
@@ -0,0 +1,125 @@
...
+	for {
+		select {
+		case <-s.done:
+			return
+		case <-s.notify:
+		}
+
+		for {
+			if current := s.dc(); current != dc {
+				dc, inFlight = current, nil
```
boks1971, line 109: this is clearing inFlight on a data channel change. 

Not sure if we want to keep checking for data channel. Maybe, it can be set and updated on change or even `newDataTrackSender` created by engine on data channel change. Would be good to keep this simple.

ladvoc: I've made some changes to data channel sender in [5bbc090](https://github.com/livekit/server-sdk-go/pull/987/commits/5bbc090fbc71d81196b03188c51b244cfee67bae) to address your comments. This clarifies that in-flight packets for a particular frame should be dropped when data channel changes, and the data channel itself is set after the sender is created. For the latter point, please let me know if this is what you had in mind.

datatracksender.go
ladvoc: Addressed in [8a03759](https://github.com/livekit/server-sdk-go/pull/987/commits/8a0375980869b982109dc7f6cc3187b5708cd682).

engine.go
```diff
@@ -557,6 +566,20 @@ func (e *RTCEngine) createPublisherPCLocked(configuration webrtc.Configuration)
...
```
cnderrauber, line 581: Maybe move the data channel configuration statements (SetBufferedAmountLowThreshold, OnBufferedAmountLow, OnOpen) into the sender.setDataChannel

cnderrauber (approved):

PR #991, opened 2026-09-03
Data track publishing

ladvoc:
Add support for publishing data tracks.

Architecture and public API closely match [Rust](https://github.com/livekit/rust-sdks/tree/main/livekit-datatrack) and [JS](https://github.com/livekit/client-sdk-js/tree/main/src/room/data-track) clients:
- The user publishes data tracks using the `LocalParticipant.PublishDataTrack` API.
- A `LocalManager` is owned by the room to manage internal state and transitions for all local data tracks. 

Areas to review:
- Some APIs meant for internal use (e.g., `LocalManager`) are currently public. Would like to understand if there is a better way to organize packages to avoid this.
- The `LocalManager` is currently owned by `LocalParticipant`, but I am not sure this is the right place for it to live.

Closes BOT-541


localdatatrack.go
```diff
@@ -0,0 +1,69 @@
...
+// PublishDataTrack publishes a data track and waits for the server to accept it. A 10 second
+// deadline applies when ctx has none. The name must be unique among the participant's data tracks.
+func (p *LocalParticipant) PublishDataTrack(ctx context.Context, name string, opts ...datatrack.PublishOption) (*datatrack.LocalTrack, error) {
+	options := datatrack.PublishOptions{Name: name}
+	for _, opt := range opts {
+		opt(&options)
+	}
+
+	_, hasDeadline := ctx.Deadline()
+	ctx, cancel := withDefaultTimeout(ctx, dataTrackPublishTimeout)
```
cnderrauber, line 57: do we still need to apply the timeout if the ctx has deadline?

cnderrauber (approved):

localdatatrack.go
ladvoc: No, `withDefaultTimeout` only applies the timeout when `ctx` has no deadline (see requestresponse.go:90-97).

PR #992, opened 2026-09-03
Data track subscribing

ladvoc:
Adds support for subscribing to data tracks.

Architecture and public API closely match [Rust](https://github.com/livekit/rust-sdks/tree/main/livekit-datatrack) and [JS](https://github.com/livekit/client-sdk-js/tree/main/src/room/data-track) clients:
- When a remote participant publishes a data track, the user receives an `OnDataTrackPublished` event to receive the track object which they can then use to subscribe if desired.
  - Multiple application-level subscriptions are supported for a single track; internally, packets are still only received and processed once and will be fanned out.
- A `RemoteManager` is owned by the room to manage internal state and transitions for all remote data tracks. 

Areas to review:
- Same as in previous PR; some APIs meant for internal use (e.g., `RemoteManager`) are currently public. Would like to understand if there is a better way to organize packages to avoid this.
- The `RemoteManager` is currently owned by `LocalParticipant`, but I am not sure this is the right place for it to live.

Closes BOT-541


room.go
```diff
@@ -369,6 +370,7 @@ func NewRoom(callback *RoomCallback) *Room {
 	r.LocalParticipant = newLocalParticipant(r.engine, r.callback, r.serverInfo, r.log)
 	r.localDataTracks = datatrack.NewLocalManager(datatrack.LocalManagerParams{Transport: localDataTrackTransport{engine: r.engine}, Logger: r.log})
 	r.LocalParticipant.dataTracks = r.localDataTracks
+	r.remoteDataTracks = datatrack.NewRemoteManager(datatrack.RemoteManagerParams{Transport: remoteDataTrackTransport{room: r}, Logger: r.log})
```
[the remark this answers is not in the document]
ladvoc: This is addressed by follow-up PRs (#993 and #994) to keep the scope of this PR limited.

datatrack/remote.go
```diff
@@ -0,0 +1,673 @@
...
+		m.params.Logger.Warnw("dropping invalid data track packet", err)
+		return
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	sid, known := m.subHandles[trackHandle(packet.Handle)]
+	track := m.descriptors[sid]
+	if !known || track == nil || track.subscription != subscriptionActive {
+		m.params.Logger.Debugw("dropping data track packet without subscription", "handle", packet.Handle)
```
cnderrauber, line 285: The packet will be dropped silently if debugging is not enabled. Should we use info or warn here?

datatrack/remote.go
```diff
@@ -0,0 +1,673 @@
...
+		return
+	}
+	// the send happens under the lock so the channel is never closed underneath it
+	select {
+	case track.packets <- packet:
+	default:
+		m.params.Logger.Debugw("dropping data track packet, pipeline is behind", "sid", sid)
```
cnderrauber: as same as above

datatrack/remote.go
```diff
@@ -0,0 +1,673 @@
...
+	manager           *RemoteManager
+	publisherIdentity string
+	unpublished       core.Fuse
+	maxPartialFrames  atomic.Int64
+
+	// streamList is an immutable snapshot of streams, read by the worker without the lock
+	streamList atomic.Pointer[[]*Stream]
+
+	// guarded by manager.mu
+	info         Info
```
cnderrauber, line 361: Using the mutex of one structure inside another structure is highly error-prone.

datatrack/remote.go
```diff
@@ -0,0 +1,673 @@
...
+	if s.closed {
+		return
+	}
+	select {
+	case s.frames <- frame:
+		return
+	default:
+	}
+	select {
+	case <-s.frames:
```
cnderrauber, line 627: It has a chance to drop a frame that can be buffered when the subscriber pop a frame from a full channel. Can we use a queue instead of the channel for the frame buffer?

datatrack/remote.go
ladvoc: That makes sense, addressed in [9b0d4c1](https://github.com/livekit/server-sdk-go/pull/992/commits/9b0d4c152c38199fc75e932ed6d916830b5cc235); please let me know if this is what you had in mind.

cnderrauber: The `RemoteManager` lock track's mutex directly now.. can we make `RemoteManager`/`RemoteTrack` use its own mutex only?

datatrack/remote.go
ladvoc: Addressed in [1ab44ed](https://github.com/livekit/server-sdk-go/pull/992/commits/1ab44edd2c9ec3ee3629fbcf923394b1d71d7e39).

datatrack/remote.go
ladvoc: Addressed in [9894af6](https://github.com/livekit/server-sdk-go/pull/992/commits/9894af66c746a510d2289b8cf03bd41edda1ed6a).

datatrack/remote.go
ladvoc: Addressed in [9894af6](https://github.com/livekit/server-sdk-go/pull/992/commits/9894af66c746a510d2289b8cf03bd41edda1ed6a).

datatrack/remote.go
ladvoc: I think I have addressed this now in [49797e4](https://github.com/livekit/server-sdk-go/pull/992/commits/49797e42dd1c551d60eb2386925619b56b84255b); please let me know if this follows best practices and if there's any additional changes you'd like to see.

PR #415, opened 2024-03-07
Accept and send DTMF via data channel

dennwc:
Accept and send DTMF via data channel. Also take the chance to make data callbacks and functions accept an interface for data messages.

Blocked on https://github.com/livekit/protocol/pull/630

PR #1001, opened 2026-09-10
Publish DTMF Helper for Local Participant

ayukumar261:
**Problem**

The JavaScript, Rust, Python, and Node SDKs have DTMF publishing helpers that always use the reliable data channel. The Go SDK has no equivalent, so developers have to build the packet themselves and remember to pass `WithDataPublishReliable(true)`.

That matters because `PublishDataPacket` has defaulted to the lossy channel since #421 (v2.1.0), and the lossy channel has been unordered since #659 (v2.8.0). Digits sent with the default can be dropped or arrive out of order.

**Changes**

- Added `PublishDTMF(code, digit)` to `LocalParticipant`. It wraps `PublishDataPacket` with `WithDataPublishReliable(true)`, so a `SipDTMF` packet always goes over the reliable channel. The helper accepts no publishing options.
- Updated a stale comment above the default in `PublishDataPacket`. It said the Go default matches the protobuf default. That was true when #415 wrote it, since both defaulted to reliable. #421 flipped the Go default to lossy but left the comment, and the protobuf zero value is still `RELIABLE`.

**Testing**

`TestPublishDTMF` publishes `1 2 3 #` and asserts that the subscriber receives four `*livekit.SipDTMF` packets with the expected codes and digits in order. It also checks the publisher’s WebRTC `GetStats` for four messages on the `_reliable` data channel and none on `_lossy`.

- The updated integration test passed against a local LiveKit server with `CGO_ENABLED=0`.
- `mage test`, which includes `-race`, passed locally before this API simplification. The latest full-suite attempt was blocked by a missing local `soxr` dependency.

**Notes**

- The helper sends DTMF digits to all participants in the room.
- Reliable delivery matches the other SDKs.

### DIFF
--- a/localparticipant.go
+++ b/localparticipant.go
@@ -663,11 +663,14 @@ func (p *LocalParticipant) PublishDataPacket(pck DataPacket, opts ...DataPublish
 // PublishDTMF sends a SIP DTMF digit to the room as a livekit.SipDTMF data packet.
 // code is the RFC 4733 event code and digit is the key it represents, e.g. 11 and "#".
 //
-// DTMF is always sent via the RELIABLE channel so that digits arrive in order,
-// see WithDataPublishDestination for choosing specific participants.
+// DTMF always uses the RELIABLE channel so digits arrive in order, overriding any
+// WithDataPublishReliable(false). See WithDataPublishDestination to target participants.
 func (p *LocalParticipant) PublishDTMF(code uint32, digit string, opts ...DataPublishOption) error {
-	opts = append(opts, WithDataPublishReliable(true))
-	return p.PublishDataPacket(&livekit.SipDTMF{Code: code, Digit: digit}, opts...)
+	// Copy opts so appending never modifies the caller's slice.
+	allOpts := make([]DataPublishOption, 0, len(opts)+1)
+	allOpts = append(allOpts, opts...)
+	allOpts = append(allOpts, WithDataPublishReliable(true))
+	return p.PublishDataPacket(&livekit.SipDTMF{Code: code, Digit: digit}, allOpts...)
 }
 
 // UnpublishTrack stops publishing a track and removes it from the room.
--- a/localparticipant.go
+++ b/localparticipant.go
@@ -659,6 +659,16 @@ func (p *LocalParticipant) PublishDataPacket(pck DataPacket, opts ...DataPublish
 	return p.engine.publishDataPacket(dataPacket, kind)
 }
 
+// PublishDTMF sends a SIP DTMF digit to the room as a livekit.SipDTMF data packet.
+// code is the RFC 4733 event code and digit is the key it represents, e.g. 11 and "#".
+//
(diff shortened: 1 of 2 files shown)


localparticipant.go
```diff
@@ -659,6 +660,16 @@ func (p *LocalParticipant) PublishDataPacket(pck DataPacket, opts ...DataPublish
...
 }
 
+// PublishDTMF sends a SIP DTMF digit to the room as a livekit.SipDTMF data packet.
+// code is the RFC 4733 event code and digit is the key it represents, e.g. 11 and "#".
+//
+// DTMF is always sent via the RELIABLE channel so that digits arrive in order,
+// see WithDataPublishDestination for choosing specific participants.
+func (p *LocalParticipant) PublishDTMF(code uint32, digit string, opts ...DataPublishOption) error {
+	opts = append(opts, WithDataPublishReliable(true))
+	return p.PublishDataPacket(&livekit.SipDTMF{Code: code, Digit: digit}, opts...)
```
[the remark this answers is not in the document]
ayukumar261: Great catch! I implemented the change in [`c77ff31`](https://github.com/livekit/server-sdk-go/pull/1001/commits/c77ff311770d3ab671100c9f5ea6bd99275e38a8) with single tweak: `allOpts` instead of `publishOpts` to match`HMACKeySet.ParseWithClaims` in livekit/protocol's [`utils/jwtutil`](https://github.com/livekit/protocol/blob/a4f4b5c0c23f/utils/jwtutil/jwtutil.go#L121).

I also considered using `slices.Clone(opts)`, which is shorter, but it would add a `slices` import to `localparticipant.go` for a single call, so I went with the explicit copy.

localparticipant.go
```diff
@@ -659,6 +660,19 @@ func (p *LocalParticipant) PublishDataPacket(pck DataPacket, opts ...DataPublish
 	return p.engine.publishDataPacket(dataPacket, kind)
 }
 
+// PublishDTMF sends a SIP DTMF digit to the room as a livekit.SipDTMF data packet.
+// code is the RFC 4733 event code and digit is the key it represents, e.g. 11 and "#".
+//
+// DTMF always uses the RELIABLE channel so digits arrive in order, overriding any
+// WithDataPublishReliable(false). See WithDataPublishDestination to target participants.
+func (p *LocalParticipant) PublishDTMF(code uint32, digit string, opts ...DataPublishOption) error {
```
cnderrauber, line 668: There are no options for publishing dtmf in other sdks, so go-sdk doesn't need options too.

ayukumar261: Thanks for reviewing! I removed the `opts` argument from the helper. 

However, I was looking at the Rust SDK `publish_dtmf` helper, and it includes `destination_identities` for destination support. If you think I should add a `destinationIdentities...` argument for the Go SDK `PublishDTMF` helper, please let me know!

cnderrauber (commented):
LGTM! Thank you for adding this.

cnderrauber (approved):

PR #1002, opened 2026-09-11
failover: retry the same host when no fallback region exists, and cover the Cloud API hosts

u9g:
## Why

A CLI call to `cloud-api.livekit.io` hung for 10 seconds and timed out on 2026-09-11 at 14:44Z. The request reached Cloudflare's edge and was acknowledged there, but no LiveKit system ever saw it: traefik's per-backend counters, the OCI load balancer's metrics, and cloud-api's request spans were all clean for that minute. Every other call in the surrounding minutes succeeded. The failure was a single lost request between Cloudflare and the origin.

The SDK's failover transport could not help, for two reasons:

- `isCloud` only matches `*.livekit.cloud`, so a `cloud-api.livekit.io` request always got exactly one attempt.
- Even with failover engaged, the loop gives up as soon as no untried region exists, which is always the case for cloud-api because it has a single origin and `/settings/regions` returns 404.

## What

Two commits, each test-first:

1. **Retry the same host when no fallback region exists.** A retryable failure with no untried region used to be surfaced after one attempt. It now retries against the same host, bounded by the existing attempt count and backoff. This matches the cross-region path, which already retries both transport errors and 5xx responses. `TestAPI_RegionDiscoveryUnreachable` still holds: a region that fails deterministically is surfaced once attempts are exhausted.
2. **Enable retries for the LiveKit Cloud API hosts.** A new `isCloudAPI` check matches `cloud-api.<env>.livekit.io` case-insensitively and joins `isCloud` in the attempts policy. Cloud API hosts skip `/settings/regions` discovery entirely, since they have a single origin and the endpoint 404s; otherwise every retry would pay the independent 2-second discovery timeout. `isCloud` itself is unchanged since it also drives region-URL parsing for project hosts.

The existing `minFailoverTimeout` gate still applies, so requests with a budget under 5 seconds get a single attempt as before.

## Behavior change

For cloud-api calls with a budget of 5 seconds or more, a lost request now costs up to three attempts with backoff. A truly dead origin takes up to three times the per-attempt budget to surface instead of one. A single lost request becomes a sub-second blip.

## Cross-SDK parity

Same change in the other server SDKs: livekit/python-sdks#811 and livekit/node-sdks#723.

## Testing

- `go test ./ -run 'Failover|PrepareContext|DialContext'` passes.
- New tests: `TestFailoverRetriesSameHostOnTransportError`, `TestFailoverRetriesSameHostOn5xx`, `TestFailoverCloudAPISkipsRegionDiscovery`, plus cloud-api rows in `TestFailoverAttempts`. Each was confirmed failing before its implementation commit.
- The root package's `integration_test.go` cases need a running LiveKit server and fail identically on `main`.

### DIFF
--- a/failover.go
+++ b/failover.go
@@ -265,7 +265,9 @@ func (t *failoverTransport) failover(req *http.Request, maxAttempts int, timeout
 		}
 		nextScheme, nextHost, ok := nextRegion(regions, tried)
 		if !ok {
-			return terminate(resp, err, cancel) // no untried region left
+			// With no fallback region, a retryable failure is retried against
+			// the same host.
+			nextScheme, nextHost = scheme, host
 		}
 
 		status := 0
--- a/failover.go
+++ b/failover.go
@@ -69,13 +69,20 @@ type failoverConfig struct {
 }
 
 // attempts returns the total request attempts for a host; 1 means no failover.
-// Failover engages only when enabled and the host is a LiveKit Cloud domain
-// (or force is set).
+// Failover engages only when enabled and the host is a LiveKit Cloud project
+// or Cloud API domain (or force is set).
 func (c failoverConfig) attempts(hostname string) int {
-	if c.enabled && (c.force || isCloud(hostname)) {
+	if c.enabled && (c.force || isCloud(hostname) || isCloudAPI(hostname)) {
 		return failoverMaxAttempts
 	}
 	return 1
+}
+
+// isCloudAPI reports whether the hostname is a LiveKit Cloud API endpoint
+// (cloud-api.livekit.io or a cloud-api.<env>.livekit.io variant).
+func isCloudAPI(hostname string) bool {
+	hostname = strings.ToLower(hostname)
+	return strings.HasPrefix(hostname, "cloud-api.") && strings.HasSuffix(hostname, ".livekit.io")
 }
 
 type failoverEnabledKey struct{}
@@ -238,6 +245,8 @@ func (t *failoverTransport) failover(req *http.Request, maxAttempts int, timeout
 	scheme, host := req.URL.Scheme, req.URL.Host
 	tried := map[string]struct{}{strings.ToLower(host): {}}
 	var regions *livekit.RegionSettings // discovered lazily on the first failure
+	// A Cloud API host has a single origin; region discovery is never consulted.
+	discover := !isCloudAPI(req.URL.Hostname())
 
 	var resp *http.Response
 	var err error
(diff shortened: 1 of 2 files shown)


failover.go
```diff
@@ -265,7 +271,12 @@ func (t *failoverTransport) failover(req *http.Request, maxAttempts int, timeout
 		}
 		nextScheme, nextHost, ok := nextRegion(regions, tried)
 		if !ok {
-			return terminate(resp, err, cancel) // no untried region left
+			// With no fallback region, a 5xx is final; only a transport error is
```
biglittlebigben: Why aren't 5xx retried?

failover.go
```diff
@@ -69,15 +69,21 @@ type failoverConfig struct {
 }
 
 // attempts returns the total request attempts for a host; 1 means no failover.
-// Failover engages only when enabled and the host is a LiveKit Cloud domain
-// (or force is set).
+// Failover engages only when enabled and the host is a LiveKit Cloud project
+// or Cloud API domain (or force is set).
 func (c failoverConfig) attempts(hostname string) int {
-	if c.enabled && (c.force || isCloud(hostname)) {
+	if c.enabled && (c.force || isCloud(hostname) || isCloudAPI(hostname)) {
```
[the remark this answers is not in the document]
biglittlebigben: Is there a way to avoid this call to regions?

u9g:
> Is there a way to avoid this call to regions?

Yes, and fixed in the latest push.

failover.go
u9g: Two reasons. First, scope: the incident this fixes was a request that got no response at all, and a transport error is the only signal that the origin never saw it.

A 5xx means the origin, or traefik in front of it, processed it and answered, so a same-host retry against the same origin is much less likely to change the outcome.

Second, safety: most twirp calls here are not idempotent, and re-sending one that the server already executed and then 500'd on is a double-execution risk.

The `X-Livekit-Request-Id` header is meant for server-side dedup, but cloud-api doesn't dedup on it today.

This also preserves the existing contract that `TestAPI_RegionDiscoveryUnreachable` pins: a 5xx with no fallback is surfaced.

If you'd rather retry the proxy-class codes, 502/503/504 from traefik do mean the request most likely never reached the app, so retrying those same-host while surfacing 500 would be a reasonable middle ground. Happy to do that in this PR if you prefer it.

biglittlebigben: I'd argue that the non idempotence issue already exists in case of transport error: the request may make its way to the server, but the response may never make it back. Unless Cloudflare synthesizes an HTTP error at the proxy level in case of timeout (before the SDK http timeout fires)?

biglittlebigben (approved):

